PR #1, opened 2022-11-04
Add initial set of tests

loganek:
We'll use them later to experiment with test harness


tests/core/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
cimacmillan, line 1: It'd be worth sticking to npm/nodejs for all the building than a mix of npm/bash. The package.json is a good self-documenter for the commands that can be run; and nodejs would be more maintainable, assuming this package will scale up. In this case it's fine, but worth settling on it early. It's something I'd be happy to add as a separate PR, if you think it's a good idea.

Eg:

*build.ts*
```
...
function getTestFiles() ...
async function compileTestFile(...) { exec('npm run asc {...}') }

Promise.all(getTestFiles().map(compileTestFile))
```

*package.json*
```
"scripts": {
    "build": "ts-node build.ts"
}
```

tests/core/testsuite/args_get-multiple-arguments.ts
```diff
@@ -0,0 +1,31 @@
+import {
+    args_sizes_get,
+    args_get,
+    errno
+} from '@assemblyscript/wasi-shim/assembly/bindings/wasi_snapshot_preview1';
```
cimacmillan, line 5: This way of writing tests looks nice. A couple of questions I had about assemblyscript,

1. Is there a delay between adding adding wasi spec, and it being included in the assembly script package? Lets say  a new WASI API is added, if assemblyscript needs to do a release before a test can be added, then maybe that introduces some churn. 
2. Does assemblyscript bundle other things into the outputted wasm like a garbage collector? It'd be interesting to know what the footprint of that is.

Either way, this looks like a good method for these core tests.

cimacmillan (approved):
LGTM!

tests/core/build.sh
loganek: Yeah, I like it. Feel free to make a PR with that change!

tests/core/testsuite/args_get-multiple-arguments.ts
loganek: > Is there a delay between adding adding wasi spec, and it being included in the assembly script package? Lets say a new WASI API is added, if assemblyscript needs to do a release before a test can be added, then maybe that introduces some churn.

I'm not concerned about that for 2 reasons:
1. We can always contribute to `wasi-shim` package
2. If 1. is too slow, we can always (temporarily) add missing declarations in this package (you can see [here](https://github.com/AssemblyScript/wasi-shim/blob/0d84fc60e493c8dc2750db5d5faffd240401e469/assembly/bindings/wasi_snapshot_preview1.ts) - it's not that complicated).

> Does assemblyscript bundle other things into the outputted wasm like a garbage collector? It'd be interesting to know what the footprint of that is.

The footprint is much less than e.g. C or rust (at least out of the box, without tuning compilation parameters for those languages):
```
$ ls -lh testsuite/*.wasm
-rw-r--r-- 1 mkolny domain^users 7.6K Nov  4 17:56 testsuite/args_get-multiple-arguments.wasm
-rw-r--r-- 1 mkolny domain^users 1.8K Nov  4 17:56 testsuite/args_sizes_get-multiple-arguments.wasm
-rw-r--r-- 1 mkolny domain^users 1.8K Nov  4 17:56 testsuite/args_sizes_get-no-arguments.wasm
-rw-r--r-- 1 mkolny domain^users 1.8K Nov  4 17:57 testsuite/fd_write-to-invalid-fd.wasm
-rw-r--r-- 1 mkolny domain^users 4.9K Nov  4 17:57 testsuite/fd_write-to-stdout.wasm
```

I'll explain and document all the choices later on (I'm planning to have `doc/` with all the things I've discussed and agreed on with people offline).

PR #4, opened 2022-11-09
Remove bash build script in favour of typescript

cimacmillan:
### What

Replaced bash build script with ts build script

### Why 

With the assumption that this package will grow in complexity, the scripts used to prepare, run tests and publish test results will also grow in complexity. I think this will be easier to maintain with a high-level language like TS or Python. Proposing to use a ts build script here, as the type annotations help with readability. These can be started with `npm run build` 

Discussed briefly in https://github.com/WebAssembly/wasi-testsuite/pull/1

### How

Added typescript and ts-node devDependencies. Start `build.ts` when the `npm run build` command is used.


tests/core/package.json
```diff
@@ -19,9 +20,12 @@
...
   "devDependencies": {
     "@assemblyscript/wasi-shim": "^0.1.0",
+    "@types/node": "^18.11.9",
     "@typescript-eslint/eslint-plugin": "^5.42.0",
     "@typescript-eslint/parser": "^5.42.0",
     "assemblyscript": "^0.22.0",
-    "eslint": "^8.26.0"
+    "eslint": "^8.26.0",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.8.4"
```
loganek, line 29: why do we need typescript dependency?

tests/core/buildscripts/build.ts
```diff
@@ -0,0 +1,16 @@
...
+async function compileTests() {
+    const testSuiteFiles = getPathsInDirectory("testsuite");
+    const testFiles = testSuiteFiles.filter(filepath => doesFileHaveExtension(filepath, "ts"));
+    const pendingCompilations = testFiles.map(filepath => compileWithAsc(filepath,  mapFileExtension(filepath, "ts", "wasm")));
+    return Promise.all(pendingCompilations);
+}
+
+compileTests()
+    .then(() => console.log("Tests compiled"))
+    .catch((e) => console.error(`Tests failed to compile: ${e}`));
```
loganek, line 16: will that script return status code !=0 when run on node? Otherwise we won't be able to notify CI script if the compilation fails.

cimacmillan: Good point, will check this

tests/core/package.json
cimacmillan: It's a dependency of ts-node, where it's JIT transpiling the ts to js

tests/core/package-lock.json
```diff
@@ -26,6 +29,18 @@
         "url": "https://opencollective.com/assemblyscript"
       }
     },
+    "node_modules/@cspotcode/source-map-support": {
+      "version": "0.8.1",
+      "resolved": "https://amazon-149122183214.d.codeartifact.us-west-2.amazonaws.com:443/npm/shared/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz",
```
loganek, line 34: Can we please use public npm endpoint?

loganek (changes requested):

tests/core/package.json
loganek: ah, I thought ts-node will pull that dependency automatically, but looks like it does not.

tests/core/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "asc": "asc",
-    "lint": "eslint . --ext .ts"
+    "lint": "eslint . --ext .ts",
+    "build": "ts-node buildscripts/build.ts"
```
loganek, line 10: I wonder if we could have some kind of ci-build or release-build that runs both linter and compiles the code, and use it in the compile-tests.yml?

PR #8, opened 2022-11-16
Rename `core` and `libc` to `assemblyscript` and `c`, respectively.

sunfishcode:
Rename `tests/core` to `tests/assemblyscript`, and `tests/libc` to `tests/c`, so that the directory names reflect the name of the source languages that the tests in each directory are written in.


loganek:
I like the idea of including language names in directory names. Before this change directory names described how tests are built and their purpose (so `core` means we're testing the API directly without going through any intermediate layers like standard library); if we we want to keep language names as part of the directory name, we might consider having names like `assemblyscript-core` etc. 

TBH I don't have strong opinion and I'm happy with the proposed change, but I'd like to know your thoughts on that.

loganek:
Sounds good; we can introduce sub-directories if needed.

PR #18, opened 2022-11-22
Fix build script for assemblyscript

loganek:
so it correctly reports compilation failures.

### DIFF
--- a/tests/assemblyscript/buildscripts/build.ts
+++ b/tests/assemblyscript/buildscripts/build.ts
@@ -20,15 +20,32 @@ async function compileTests() {
     const testSuiteFiles = getPathsInDirectory("testsuite");
     const pendingCompilations = testSuiteFiles
         .filter(filePath => doesFileHaveExtension(filePath, "ts"))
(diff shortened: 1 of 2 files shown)


tests/assemblyscript/buildscripts/build.ts
```diff
@@ -20,15 +20,32 @@ async function compileTests() {
...
+    const errors = (await Promise.allSettled(pendingCompilations))
+        .map(p => {
+            if (p.status === 'fulfilled' && p.value !== undefined) {
+                return `Failed to compile ${p.value.source}:\n${p.value.error}`;
+            } else if (p.status === 'rejected') {
+                return `Execution failed:\n${p.reason}`;
+            } else {
+                return undefined;
+            }
+        }).filter(p => p !== undefined).map(p => p as string)
```
cimacmillan, line 38: np. I think this last map doesn't do anything as types are stripped on execution. I think it could be replaced by:
```
const errors = ...
        }).filter(p => p !== undefined) as string[]
```

cimacmillan (approved):
LGTM, sorry for missing this!

PR #26, opened 2022-11-28
Add pread/pwrite tests to libc testsuite

loganek:


tests/c/testsuite/pwrite-with-access.c
```diff
@@ -0,0 +1,48 @@
...
+
+  ret = read(fd, read_buf, sizeof(read_buf));
+  assert(ret == strlen(full_content));
+
+  assert(strncmp("vertestng text", read_buf, strlen(full_content)) == 0);
+
+  ret = close(fd);
+  assert(ret == 0);
+
+  ret = remove(test_file);
```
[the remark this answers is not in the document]
loganek: yeah, I've added a check at the beginning of the file so at least the test fails with a meaningful reason; however, I reconsidered that and added a feature in the test runner to automatically cleanup files and directories with `.cleanup` suffix. I also updated the documentation.

tests/c/testsuite/pwrite-with-access.c
```diff
@@ -0,0 +1,48 @@
...
+
+  /* make sure the test file doesn't exist */
+  ret = access(test_file, F_OK);
+  assert(ret != 0);
+
+  fd = open(test_file, O_WRONLY | O_CREAT);
+  assert(fd > 0);
+
+  ret = pwrite(fd, full_content, strlen(full_content), 0);
+  assert(ret == strlen(full_content));
```
sunfishcode, line 25: This assert and the others like it are stricter than what POSIX requires, and I believe WASI wants to follow POSIX here. In particular, `pwrite` is permitted to write fewer bytes than requested, and still succeed.

loganek: good point, updated the code.

sunfishcode (approved):

Issue #24, opened 2022-11-25
Migrate Rust tests from wasmtime


loganek:
Looks like wasmtime already has a good coverage for WASI interface [here](https://github.com/bytecodealliance/wasmtime/tree/main/crates/test-programs/wasi-tests). It'd be beneficial to move those tests here so other runtimes can easily access them

yamt:
FYI, some of them are testing behaviors which i couldn't find the authoritive text in wasi.
they might be implementation (wasmtime) specific.

* https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L640
* https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L1382-L1383
* https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L1669-L1671
* https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L2319-L2334
* https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/test/wasmtime-wasi-tests-blacklist-Linux.txt#L1

sunfishcode:
> FYI, some of them are testing behaviors which i couldn't find the authoritive text in wasi. they might be implementation (wasmtime) specific.
> 

>     * https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L640

>                 /* Note: wasmtime returns EINVAL for embedded NULs */

This error condition does not occur in POSIX so there is no obvious choice here. Do you have a suggestion for an error code that makes more sense than `EINVAL` here?

>     * https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L1382-L1383

>              /*
                 * Note: wasmtime directory_seek.rs test expects EBADF.
                 * Why not EISDIR?
                 */

No strong opinion here; `EISDIR` sounds reasonable; I'll look into changing Wasmtime.

>     * https://github.com/yamt/toywasm/blob/cf6f19d7b2921c7a244e95e38f36c5b6339a708d/lib/wasi.c#L1669-L1671

[...]

PR #32, opened 2022-12-08
Copy some of the WASI Rust tests from wasmtime

loganek:


yamt:
why not all of them?
what was the criteria?

loganek:
For now I only migrated tests starting from A to F; haven't done all of them to get early feedback before I spend more time on that.

yamt:
> For now I only migrated tests starting from A to F; haven't done all of them to get early feedback before I spend more time on that.

ok. it makes sense.

i have concerns on some of test cases: https://github.com/WebAssembly/wasi-testsuite/issues/24#issuecomment-1345861497

loganek:
> > For now I only migrated tests starting from A to F; haven't done all of them to get early feedback before I spend more time on that.
> 
> ok. it makes sense.
> 
> i have concerns on some of test cases: [#24 (comment)](https://github.com/WebAssembly/wasi-testsuite/issues/24#issuecomment-1345861497)

Thanks, I've taken the ones related to this PR into account.

tests/rust/src/bin/directory_seek.rs
```diff
@@ -27,7 +27,7 @@ unsafe fn test_directory_seek(dir_fd: wasi::Fd) {
         wasi::fd_seek(fd, 0, wasi::WHENCE_CUR)
             .expect_err("seek on a directory")
             .raw_error(),
-        wasi::ERRNO_BADF
+        wasi::ERRNO_ISDIR
```
yamt, line 30: i guess it's more reasonable to accept both.

loganek: I'd rather have tests very specific so implementers can use it as a guidance. IIRC @sunfishcode will look into changing that in wasmtime, do you know of any other runtime where the error code for this case is `BADF`?

yamt: > I'd rather have tests very specific so implementers can use it as a guidance.

without having it in the spec itself? why?

> IIRC @sunfishcode will look into changing that in wasmtime, do you know of any other runtime where the error code for this case is `BADF`?

toywasm. (because it followed wasmtime)

loganek: > without having it in the spec itself? why?

There's no error code mentioned in spec itself, so are you suggesting we should just check if there's **any** error, and ignore its value? I don't mind doing it either way, but I think even if the spec doesn't mention the value of the error code, we can still have a specific assertion here, and use it as a guidance for the future spec changes.

yamt: my suggestion is
* accept both values here for now (not necessarily "any error". probably just BADF and ISDIR are enough.)
* start mentioning error codes in the spec

loganek: ok, I updated it so it accepts BADF, ISDIR and NOTCAPABLE (this one is returned by WAMR).

loganek:
@sunfishcode / @yamt let me know if you have any more comments, otherwise I'll go ahead and merge that (we can always update the code if needed).

tests/rust/src/bin/directory_seek.rs
yamt: although NOTCAPABLE sounds very awkward to me, i guess it isn't a problem in this PR

yamt (approved):

sunfishcode:
Looks good to me; we can iterate from here.

I also agree that `ENOTCAPABLE` is awkward. WASI inherited it from CloudABI, and within WASI at least, it hasn't turned out to be useful, because most application code isn't expecting it. So I think we want to migrate away from using it in general. But it's fine to accept here for now.

PR #33, opened 2022-12-10
Don't fail the workflow if test execution fails

loganek:


loganek:
Merging to unblock nightly test execution

Issue #29, opened 2022-12-01
Enable test filters


loganek:
It'd be great to have flags in the test runner that allows to filter tests that are being executed.

vados-cosmonic:
Filtering tests is now possible via the reworked `buck2`-based build system.

Going to close this for now, happy to reopen if I'm missing something!

PR #35, opened 2022-12-19
Implement a simple test filter

yamt:
cf. https://github.com/WebAssembly/wasi-testsuite/issues/29

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,13 +38,13 @@ python3 test-runner/wasi_test_runner.py                                         
     -r adapters/wasmtime.sh # path to a runtime adapter
 ```
 
-Optionally you can specify test cases to skip with the `--filter` option.
+Optionally you can specify test cases to skip with the `--exclude-filter` option.
 
 ```bash
 python3 test-runner/wasi_test_runner.py                                                  \
     -t ./tests/assemblyscript/testsuite/ `# path to folders containing .wasm test files` \
        ./tests/c/testsuite/                                                              \
-    --filter examples/skip.json                                                          \
+    --exclude-filter examples/skip.json                                                  \
     -r adapters/wasmtime.sh # path to a runtime adapter
 ```
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,16 @@ python3 test-runner/wasi_test_runner.py                                         
     -r adapters/wasmtime.sh # path to a runtime adapter
 ```
 
+Optionally you can specify test cases to skip with the `--filter` option.
+
+```bash
+python3 test-runner/wasi_test_runner.py                                                  \
+    -t ./tests/assemblyscript/testsuite/ `# path to folders containing .wasm test files` \
+       ./tests/c/testsuite/                                                              \
+    --filter examples/skip.json                                                          \
+    -r adapters/wasmtime.sh # path to a runtime adapter
+```
+
 ## Contributing
 
 All contributions are very welcome. Contributors can help with:
--- a/examples/skip.json
+++ b/examples/skip.json
@@ -1,0 +1,5 @@
+{
+    "WASI C tests": {
+        "stat-dev-ino": "d_ino is known broken in this combination of engine and wasi-sdk version."
+    }
+}
--- a/test-runner/tests/test_test_suite_runner.py
+++ b/test-runner/tests/test_test_suite_runner.py
@@ -62,8 +62,12 @@ def test_runner_end_to_end() -> None:
 
     reporters = [Mock(), Mock()]
 
+    filt = Mock()
+    filt.should_skip.return_value = (False, None)
+    filters = [filt]
+
     with patch("glob.glob", return_value=test_paths):
-        suite = tsr.run_tests_from_test_suite("my-path", runtime, validators, reporters)  # type: ignore
+        suite = tsr.run_tests_from_test_suite("my-path", runtime, validators, reporters, filters)  # type: ignore
 
     # Assert manifest was read correctly
     assert suite.name == "test-suite"
@@ -91,9 +95,15 @@ def test_runner_end_to_end() -> None:
         for config, output in zip(expected_config, outputs):
             validator.assert_any_call(config, output)
 
+    # Assert filter calls
+    for filt in filters:
+        assert filt.should_skip.call_count == 3
+        for test_case in expected_test_cases:
+            filt.should_skip.assert_any_call(suite.name, test_case.name)
+
 
 @patch("os.path.exists", Mock(return_value=False))
 def test_runner_should_use_path_for_name_if_manifest_does_not_exist() -> None:
-    suite = tsr.run_tests_from_test_suite("my-path", Mock(), [], [])
+    suite = tsr.run_tests_from_test_suite("my-path", Mock(), [], [], [])
 
     assert suite.name == "my-path"
--- a/test-runner/wasi_test_runner/filters.py
+++ b/test-runner/wasi_test_runner/filters.py
@@ -11,7 +11,7 @@ class TestFilter(ABC):
         pass
 
 
-class JSONTestFilter(TestFilter):
+class JSONTestExcludeFilter(TestFilter):
     def __init__(self, filename: str) -> None:
         with open(filename, encoding="utf-8") as file:
             self.filter_dict = json.load(file)
--- a/test-runner/wasi_test_runner/filters.py
+++ b/test-runner/wasi_test_runner/filters.py
@@ -1,0 +1,24 @@
+from typing import Tuple, Any
+from abc import ABC
+from abc import abstractmethod
+
+import json
+
+
+class TestFilter(ABC):
+    @abstractmethod
+    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
+        pass
+
+
+class JSONTestFilter(TestFilter):
+    def __init__(self, filename: str) -> None:
+        with open(filename, encoding="utf-8") as file:
+            self.filter_dict = json.load(file)
+
+    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
+        test_suite_filter = self.filter_dict.get(test_suite_name)
+        if test_suite_filter is None:
+            return False, None
+        why = test_suite_filter.get(test_name)
+        return why is not None, why
--- a/test-runner/wasi_test_runner/filters.py
+++ b/test-runner/wasi_test_runner/filters.py
@@ -1,4 +1,4 @@
-from typing import Tuple, Any
+from typing import Tuple, Union, Literal
 from abc import ABC
 from abc import abstractmethod
 
@@ -7,7 +7,9 @@ import json
 
 class TestFilter(ABC):
     @abstractmethod
-    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
+    def should_skip(
+        self, test_suite_name: str, test_name: str
+    ) -> Union[Tuple[Literal[True], str], Tuple[Literal[False], Literal[None]]]:
         pass
 
 
@@ -16,9 +18,13 @@ class JSONTestExcludeFilter(TestFilter):
         with open(filename, encoding="utf-8") as file:
             self.filter_dict = json.load(file)
 
-    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
+    def should_skip(
+        self, test_suite_name: str, test_name: str
+    ) -> Union[Tuple[Literal[True], str], Tuple[Literal[False], Literal[None]]]:
         test_suite_filter = self.filter_dict.get(test_suite_name)
         if test_suite_filter is None:
             return False, None
         why = test_suite_filter.get(test_name)
-        return why is not None, why
+        if why is not None:
+            return True, why
+        return False, None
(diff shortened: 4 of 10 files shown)


README.md
```diff
@@ -38,6 +38,16 @@ python3 test-runner/wasi_test_runner.py
     -r adapters/wasmtime.sh # path to a runtime adapter
 ```
 
+Optianally you can specify test cases to skip with the `--filter` option.
```
loganek, line 41:
```suggestion
Optionally you can specify test cases to skip with the `--filter` option.
```

README.md
```diff
@@ -38,6 +38,16 @@ python3 test-runner/wasi_test_runner.py
...
+
+```bash
+python3 test-runner/wasi_test_runner.py                                                  \
+    -t ./tests/assemblyscript/testsuite/ `# path to folders containing .wasm test files` \
+       ./tests/c/testsuite/                                                              \
+    --filter examples/skip.json
```
loganek, line 47:
```suggestion
    --filter examples/skip.json                                                          \
```

test-runner/wasi_test_runner/reporters/console.py
```diff
@@ -54,6 +54,7 @@ def finalize(self, version: RuntimeVersion) -> None:
             print(f"  Total: {suite.test_count}")
             self._print_pass(f"  Passed:  {suite.pass_count}")
             self._print_fail(f"  Failed:  {suite.fail_count}")
+            print(f"  Skipped: {suite.skip_count}")
```
loganek, line 57:
```suggestion
            self._print_skip(f"  Skipped: {suite.skip_count}")
```

test-runner/tests/test_test_suite_runner.py
```diff
@@ -62,8 +62,12 @@ def test_runner_end_to_end() -> None:
 
     reporters = [Mock(), Mock()]
 
+    filt = Mock()
+    filt.should_skip.return_value = (False, None)
+    filters = [filt]
+
     with patch("glob.glob", return_value=test_paths):
-        suite = tsr.run_tests_from_test_suite("my-path", runtime, validators, reporters)  # type: ignore
+        suite = tsr.run_tests_from_test_suite("my-path", runtime, validators, reporters, filters)  # type: ignore
```
loganek, line 70: should we validate if filters actually were called?

examples/skip.json
```diff
@@ -0,0 +1,5 @@
+{
```
loganek, line 1: I'm not sure if we actually need such a detailed filters for now. Initially I was thinking of the following requirements:
* being able to do both: include **and** exclude tests (e.g. `--filter-include ".*thread.*" --filter-exclude ".*open.*"`)
* pass regular expressions so we can e.g. disable the whole testsuite (something like `--filter-exclude "WASI C tests.*"`)

But I'm curious what are your thoughts on that.

loganek (changes requested):

examples/skip.json
yamt: my intention is to implement something simplest first. (regex etc is more complex and advanced in my taste.)

also i think it makes sense to have a filter in a file because i guess it's almost static for a given runtime.

README.md
yamt: done

README.md
yamt: done

test-runner/wasi_test_runner/reporters/console.py
yamt: done

test-runner/tests/test_test_suite_runner.py
yamt: done

examples/skip.json
loganek: Sure, I don't suggest implementing all that at once, but I'd like us to think a bit about the API before we move forward. The proposed API of the JSON file only allows for disabling tests, and I'm afraid that if we want to add more features (like the ones mentioned above) we'll either have to make backward-incompatible changes or make the API inconsistent. As I said, we don't have to implement everything in the first iteration, but at least having placeholders would probably simplify further development.

yamt: i feel compatibility concern is not too important at this point.

also, your requirements are not clear to me.
if you explain it a bit more, maybe i can suggest a schema.

loganek: Sure; so first of all, I think we should be able to allow users to both select exclusion and inclusion filter (at the moment only the exclusion is possible). Also, we might allow for wildcard/regex, but I assume this could be just a feature request for the future, and we can implement it in the future; just an idea (although feel free to come up with something different):
```json
{
  "include": [], // if empty, all tests are included
  "exclude/ignore(whatever sounds better)": [] // if empty, none of the tests are excluded
}
```
Element in the array could be a full path to a test, e.g. `WASI C tests.clock_getres-monotonic`, so we can in the future implement wildcards to only enable clock_getres tests, with: `WASI C tests.clock_getres*`.

As I said, this is just an example schema to explain what I mean; please let me know if it's clear what we try to achieve here and share your ideas.

yamt: > Sure; so first of all, I think we should be able to allow users to both select exclusion and inclusion filter (at the moment only the exclusion is possible). Also, we might allow for wildcard/regex, but I assume this could be just a feature request for the future, and we can implement it in the future; just an idea (although feel free to come up with something different):
> 
> ```json
> {
>   "include": [], // if empty, all tests are included
>   "exclude/ignore(whatever sounds better)": [] // if empty, none of the tests are excluded
> }
> ```

if a test is included in both of "include" and "exclude", "exclude" wins, right?

> 
> Element in the array could be a full path to a test, e.g. `WASI C tests.clock_getres-monotonic`, so we can in the future implement wildcards to only enable clock_getres tests, with: `WASI C tests.clock_getres*`.

while regex might be convenient for some cases, to enable/disable a set of tests,
it might be simpler to have "feature" flags in each tests.
eg. "require": ["wasi-threads"]

> 
> As I said, this is just an example schema to explain what I mean; please let me know if it's clear what we try to achieve here and share your ideas.

loganek: > if a test is included in both of "include" and "exclude", "exclude" wins, right?

I don't have preference about that; maybe worth checking what other frameworks do for consistency?

> while regex might be convenient for some cases, to enable/disable a set of tests,
> it might be simpler to have "feature" flags in each tests.
> eg. "require": ["wasi-threads"]
I think we can have both; I also like the idea of having `feature`/`tags` flags; in fact, I had similar idea to track what wasi functions are tested in a given test (see https://github.com/WebAssembly/wasi-testsuite/blob/6d4b57497ceacd304e35a7ab03dd9bb65950bd79/test-runner/wasi_test_runner/test_case.py#L36), maybe we can re-purpose this field to be a `feature`/`tag` field?

yamt: my suggestion for this PR:

* rename `--filter` option to `--exclude-filter`
* no changes to schema
* we can consider adding `--include-filter`, `--exclude-regex-filter` etc later when/if necessary.

loganek: > rename `--filter` option to `--exclude-filter`
Given we already define filters in the JSON file, I don't know if there's a need for having two cli parameters, or instead could we have just one (`--filter`) and update the schema so it allows for specifying both inclusion and exclusion - I don't have a strong opinion; just a thought, as I haven't seen them separate in other farameworks. 

> no changes to schema
Not sure if we actually need a nested structure here (testsuite->tests); if we implement regex/glob we could probably just have a list of strings? We can start with this one as it might seem a bit simpler but once we update it, we'll have to update all the consumers as well (not sure how many runtimes will onboard by then though).

I'm also not 100% sure about file vs cli parameter. Whereas I see the benefit you mentioned above, I think it 
doesn't matter that much for runtimes as they'll have their own CI scripts that wrap around the call to the tests. The benefit of having everything through CLI is that developers can easily just filter some of the tests they're currently working on without modifying file (and risking commiting it accidentally). Similar to my previous point, it is something we can change in the future, but depending on the number of adopters, it might be easier or harder.

yamt: > > rename `--filter` option to `--exclude-filter`
> 
> Given we already define filters in the JSON file, I don't know if there's a need for having two cli parameters, or instead could we have just one (`--filter`) and update the schema so it allows for specifying both inclusion and exclusion - I don't have a strong opinion; just a thought, as I haven't seen them separate in other farameworks.

given the current cli options which allows to specify multiple testsuites,
i thought it's natural to accept multiple filters.

> 
> > no changes to schema
> 
> Not sure if we actually need a nested structure here (testsuite->tests); if we implement regex/glob we could probably just have a list of strings? We can start with this one as it might seem a bit simpler but once we update it, we'll have to update all the consumers as well (not sure how many runtimes will onboard by then though).

i'm not the person who invented the testsuite->tests structure in this repo.
honestly speaking i don't think the testsuite concept makes much sense.

maybe we can use some kind of string match with test id, where test id is:
```
test_id = testsuite + '/' + test
```

> 
> I'm also not 100% sure about file vs cli parameter. Whereas I see the benefit you mentioned above, I think it doesn't matter that much for runtimes as they'll have their own CI scripts that wrap around the call to the tests. The benefit of having everything through CLI is that developers can easily just filter some of the tests they're currently working on without modifying file (and risking commiting it accidentally). Similar to my previous point, it is something we can change in the future, but depending on the number of adopters, it might be easier or harder.

as a developer, i feel it easier to copy/modify a file than tweaking cli options.
but i can understand you have a different preference.

loganek: To unblock the change, let's move on with your suggestion here https://github.com/WebAssembly/wasi-testsuite/pull/35#discussion_r1063327236, we'll adapt it once we have more feedback.

yamt: > To unblock the change, let's move on with your suggestion here [#35 (comment)](https://github.com/WebAssembly/wasi-testsuite/pull/35#discussion_r1063327236), we'll adapt it once we have more feedback.

done

test-runner/wasi_test_runner/filters.py
```diff
@@ -0,0 +1,24 @@
...
+    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
+        pass
+
+
+class JSONTestExcludeFilter(TestFilter):
+    def __init__(self, filename: str) -> None:
+        with open(filename, encoding="utf-8") as file:
+            self.filter_dict = json.load(file)
+
+    def should_skip(self, test_suite_name: str, test_name: str) -> Tuple[bool, Any]:
```
loganek, line 19: Should that be `str` instead of `Any`?

test-runner/wasi_test_runner/test_suite_runner.py
```diff
@@ -25,28 +26,52 @@ def run_tests_from_test_suite(
...
 
 
+def _skip_single_test(
+    _runtime: RuntimeAdapter, _validators: List[Validator], test_path: str
+) -> TestCase:
+    config = _read_test_config(test_path)
+    return TestCase(
+        name=os.path.splitext(os.path.basename(test_path))[0],
+        config=config,
+        result=Result(output=Output(0, "", ""), is_executed=False, failures=[]),
```
loganek, line 70: Should the `output` be optional? Or rather have an union of `output|skip_reason` and the value is picked up based on `is_executed`

test-runner/wasi_test_runner/filters.py
yamt: i made it Any because it can be None.

loganek: In that case it should be `Tuple[bool, Optional[str]]`

test-runner/wasi_test_runner/test_suite_runner.py
yamt: maybe. i'd like to postpone it for later PRs.

it would be nicer to be able to represent "timed out" as well. https://github.com/WebAssembly/wasi-testsuite/issues/42

test-runner/wasi_test_runner/filters.py
yamt: iirc such complex types are rejected by tools i was using when writing this code. i guess it's better to avoid being too eager for now.

loganek: What are those tools? I think `mypy` should be able to handle that. Unless we really except passing here something other than `str` or `None`, I'd rather stick to precise types. That also helps understanding the code; in fact, I wasn't sure what is being returned here until I've looked to the implementation.

loganek: We already have a few static code analyzers as part of CI (including mypy); I'd suggest giving it a try, and if for some reason it fails, we can open an issue for mypy and switch back to `Any` for now.

yamt: i tried more strict annotation and for some reasons it worked today.

maybe the annotation which didn't work was different from this version. i don't remember.

test-runner/wasi_test_runner/test_suite_runner.py
loganek: Alright, we can refactor that piece as part of #42 then.

test-runner/tests/test_test_suite_runner.py
loganek: it'd be also good to validate the case when the filter returns `False` and in that case the test is skipped

loganek (approved):
LGTM with a minor comment regarding tests

yamt:
@loganek are you waiting for fixes for your "minor comment"?

loganek:
I'm ok with merging that as it is but I'd like to see a bit more tests. Let me know if you can add them, otherwise I'll open an issue and might pick that up later.

PR #38, opened 2022-12-21
proc_exit-failure: use a bit more awkward exit code

yamt:
The exit code 1 is too generic:

* many of platforms have EXIT_FAILURE == 1.

* many of wasm engines exit with 1 on a unreachable trap. eg. wasm3 iwasm (wasm-micro-runtime) wasmer toywasm wasm-interp (wabt) wasmedge

  Note: wasmtime using 134 is an exception.

Use a bit more awkward value instead to avoid succeeding with a luck.


loganek (approved):
LGTM

PR #46, opened 2023-01-20
Allow overriding executable in adapters

loganek:


adapters/wasm-micro-runtime.sh
```diff
@@ -9,7 +9,7 @@ ARGS=()
 DIR=()
 ENV=()
 
-IWASM=iwasm
+IWASM="${TEST_RUNTIME_EXE:=iwasm}"
```
yamt, line 12: i think `:-` is more appropriate for this usage

adapters/wasmtime.sh
```diff
@@ -4,6 +4,8 @@ TEST_FILE=
 ARGS=()
 PROG_ARGS=()
 
+WASMTIME="${WASI_RUNTIME_EXE:=wasmtime}"
```
yamt, line 7: ditto

yamt (approved):

PR #54, opened 2023-02-03
Enable Rust tests in CI

### LINKED ISSUE #49: Add rust tests to README and testsuite-base branch
Looks like rust tests have been added, but instructions don't include them (should they?) also, the wasm isn't uploaded to the testsuite-base branch. Is this intentional or just a TODO?

https://github.com/WebAssembly/wasi-testsuite/tree/prod/testsuite-base/tests/rust

loganek:
Fixes #49

### DIFF
--- a/.github/workflows/compile-tests.yml
+++ b/.github/workflows/compile-tests.yml
@@ -43,6 +43,32 @@ jobs:
           path: tests/assemblyscript/testsuite
           if-no-files-found: error
 
+  build_rust:
+    name: Build Rust tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v3
+
+      - name: Install wasm32-wasi target for Rust
+        run: rustup target add wasm32-wasi
+
+      - name: Build tests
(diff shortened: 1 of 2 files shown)


codefromthecrypt:
I think this should be updated and triggered right? https://github.com/WebAssembly/wasi-testsuite/blob/2b74496ff41644584f6cb71f339349d27350e087/.github/workflows/daily-runtime-validation.yml

main thing is there are new tests that some runtimes might not pass due to how they are defined.

cc @waynz0r @yamt I think you are the only other two on github using this suite. heads up as rust tests from wasmtime exist now.

loganek:
@codefromthecrypt this job for now is not really used (in the future we're planning to publish results in human-readable form for potential users).
Regarding tests that runtimes don't pass - yes, I'm aware of that, and planning to review all the test cases soon to make sure we only test what's specified in the WASI spec. Users have option to skip tests if they want to pull latest version of `prod/testsuite-base` which hopefully is sufficient for now.

codefromthecrypt:
@loganek thanks for the feedback and glad there's a review vs reality of runtimes planned in the future. makes me a lot less nervous as we know people point to charts as soon as they exist, often judgmentally.

mathetake:
Kindly ping - Can we land this so that we could test without rebuilding by ourselves?

PR #55, opened 2023-02-16
Adds wazero in daily-runtime-validation

mathetake:


.github/workflows/daily-runtime-validation.yml
```diff
@@ -53,6 +59,7 @@ jobs:
           python3 test-runner/wasi_test_runner.py \
             -r ./adapters/${{ matrix.runtime }}.py \
             --json-output-location results.json \
+            -f ./tests/rust/${{ matrix.runtime }}_skip.json \
```
mathetake, line 62: I guess this causes failure until this is synced with `prod/testsuite-base` branch. Could you suggest what I should do?  @loganek

.github/workflows/daily-runtime-validation.yml
```diff
@@ -53,6 +59,7 @@ jobs:
           python3 test-runner/wasi_test_runner.py \
             -r ./adapters/${{ matrix.runtime }}.py \
             --json-output-location results.json \
+            -f ./adapters/${{ matrix.runtime }}_skip.json \
```
loganek, line 62: I wonder if that's something we can set conditionally instead of creating empty `skip` files for other adapters (e.g. set `EXTRA_RUNNER_PARAMS` with default to be empty)?

Also, my preference would be to actually run all the tests in the daily runtime validation - it's fine if that fails, it is only used for reporting, so I'd rather keep all of them running.

mathetake: got it, will change in a bit

mathetake: done. https://github.com/WebAssembly/wasi-testsuite/pull/55/commits/20bc89e67e7a253ac0b18ec3c3838444da4e959e

yamt:
i feel it makes more sense to have this in wazero repo.
(ditto for wasmtime and wamr adapters.)

mathetake:
totally I agree. if that's the general direction of this repo, I would be happy to withdraw.

loganek:
I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.

mathetake:
Cool, thank you for clarification!

yamt:
> I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.

IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.

loganek:
> > I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.
> 
> IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.

We're not advertising any runtimes. We're open to onboard any new runtimes that are ready for wide adoption (and based on github page, wazero is one of them).

yamt:
> > > I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.
> > 
> > 
> > IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.
> 
> We're not advertising any runtimes. We're open to onboard any new runtimes that are ready for wide adoption (and based on github page, wazero is one of them).

do you mean to set some criteria for a runtime to be "ready for wide adoption"?
or, do you simply mean to be open to any runtimes?

.github/workflows/daily-runtime-validation.yml
yamt: > Also, my preference would be to actually run all the tests in the daily runtime validation - it's fine if that fails, it is only used for reporting, so I'd rather keep all of them running.

i tend to disagree. always failing tests will be ignored by people. it's better to have a skip list and/or expected-fail list.

loganek: The daily job is to generate a list of runtimes and their wasi readiness. The results are not being used currently at all, but once we update all the tests so they're 1:1 with WASI spec we'll start publishing report to a website. 
Runtimes, if they choose to use wasi testsuite in the CI, can (and should, for expected failures until they're fixed) use exclusion list.

yamt: what do you mean by 1:1?

> we'll start publishing report to a website.

i feel it's simpler/better to let each runtimes publish their results by themselves if they want.

codefromthecrypt: we added code to add the dot, dot-dot at the cost of an extra stat for all users. We no longer require the skip

codefromthecrypt:
I just rebased this with the workaround that trades some performance for dot-dot compliance. What's the current thinking on this? should we plan to have a copy here or not?

My 2p is that when various runtimes are here, it becomes more obvious the impact to new tests without a fan-out or late feedback. That said, it is your repo afterall and there are other valid opinions.

.github/workflows/daily-runtime-validation.yml
codefromthecrypt: actually, it is more expensive than this, as all files end up needing inodes. I'll add a CLI arg to wazero as it is very expensive to comply with this and few will need it.

codefromthecrypt:
ok wazero now passes 100pct on darwin, linux and windows with no exceptions. Let us now if you plan to merge this or prefer to not.

loganek:
Thanks a lot for the work. Would there be an option to move the adapter to wazero, but keep wazero part of the daily test run?

codefromthecrypt:
@loganek oh sorry maybe I misunderstood. you mean changing the workflow here to check out wazero's copy of the adapter instead of hosting it here? yeah I'm sure that's fine.

yamt:
> > > > I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.
> > > 
> > > 
> > > IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.
> > 
> > 
> > We're not advertising any runtimes. We're open to onboard any new runtimes that are ready for wide adoption (and based on github page, wazero is one of them).
> 
> do you mean to set some criteria for a runtime to be "ready for wide adoption"? or, do you simply mean to be open to any runtimes?

@loganek 
i'm waiting for your answer.

loganek:
> @loganek oh sorry maybe I misunderstood. you mean changing the workflow here to check out wazero's copy of the adapter instead of hosting it here? yeah I'm sure that's fine.

Yes, exactly that, thanks.

loganek:
> > > > > I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.
> > > > 
> > > > 
> > > > IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.
> > > 
> > > 
> > > We're not advertising any runtimes. We're open to onboard any new runtimes that are ready for wide adoption (and based on github page, wazero is one of them).
> > 
> > 
> > do you mean to set some criteria for a runtime to be "ready for wide adoption"? or, do you simply mean to be open to any runtimes?
> 
> @loganek i'm waiting for your answer.

Oh sorry, I missed your comment. There are no criteria as of now. The whole point of having this daily job is to eventually generate a website with test results (what APIs are supported by which runtime) so users can go and check the status (we don't have the website yet though). I'm not sure if long-term this daily check will be hosted in this repository, but for now this is the only place we can keep it.
I'd recommend not having there runtimes that are somehow experimental, not actively maintained or not production ready though.

mathetake:
resolved conflict. good to go I guess. @loganek

codefromthecrypt:
@mathetake sorry he meant to drop wazero.py from this PR and instead check it out from wazero's via git.

mathetake:
ah ok

mathetake:
done!

yamt:
> > > > > > I don't mind keeping the adapter here, but I agree for maintenance reasons it might be easier if it was kept in wazero repository. Having said that, I'd like wazero to be included in the daily job execution, so I'd keep at least that part in the PR.
> > > > > 
> > > > > 
> > > > > IMO, wasi-testsuite should not "prefer"/"advertise" a particular set of runtimes.
> > > > 
> > > > 
> > > > We're not advertising any runtimes. We're open to onboard any new runtimes that are ready for wide adoption (and based on github page, wazero is one of them).
> > > 
> > > 
> > > do you mean to set some criteria for a runtime to be "ready for wide adoption"? or, do you simply mean to be open to any runtimes?
> > 
> > 
> > @loganek i'm waiting for your answer.
> 
> Oh sorry, I missed your comment. There are no criteria as of now. The whole point of having this daily job is to eventually generate a website with test results (what APIs are supported by which runtime) so users can go and check the status (we don't have the website yet though). I'm not sure if long-term this daily check will be hosted in this repository, but for now this is the only place we can keep it. I'd recommend not having there runtimes that are somehow experimental, not actively maintained or not production ready though.

i'm against to serve such a website.
for similar reasons why sightglass doesn't aim to be a general purpose benchmark:
https://github.com/bytecodealliance/sightglass/#this-is-not-a-general-purpose-webassembly-benchmark-suite

codefromthecrypt:
@yamt @loganek 
> i'm against to serve such a website.
for similar reasons why sightglass doesn't aim to be a general purpose benchmark:
https://github.com/bytecodealliance/sightglass/#this-is-not-a-general-purpose-webassembly-benchmark-suite

TL;DR; This isn't a benchmark, and I agree this topic is polarizing. However, I think we shouldn't try to block knowledge that runtimes pass this, and it is a good thing that bytecodealliance are considering runtimes outside their scope in WASI, just as they are at the core level. Whether that's published on a site or not is a really separate concern, which I'm ok with but understand reasons not. Since they are separate concerns, I would like them to remain as such, and allow wazero and wasmedge and anything else stable to opt-in to being in this yaml file. Website can and should be kicked to a different issue.

----

To elaborate, I agree it is relevant to discuss the future of this data even if it is only limited to bytecodealliance runtimes. However, I think they are separate steps.

### WASI caring about other runtimes is a *good thing*

wazero has a valid concern that without shared stake it is very easy for these tests to break. At least if a runtime is executed, the maintenance team can know this. You can also see other runtimes in this problem, such as the recent thing in wasmedge #64. When you don't have a convenient place for a spec to make sure they aren't breaking impls, it becomes very hard to get the point across. Having the ability for a click to figure out out the new status quo without a burden. I think it is reasonable that @loganek asked if wazero planned to run our own adapter. I think that is a good check and balance, as it prevents this repo from housing an unused and likely to break adapter.

### this is not a benchmark, this is a part of a specification and conformance to that.

Like it or not, and I include myself in this, the webassembly groups do publish sites about comformance to specs not everyone agrees with. For example, the roadmap has troubled me in the past due to how it makes runtimes that are missing or not implementing experiments seem behind. https://webassembly.org/roadmap/

The data right now is about snapshot01 which is widely used and many people including myself have asked for spec tests in the last years. Finally, they are doing it, and this gives a chance for people to know if something is actually passing or not, vs what happened in the past. I won't call out, but I have seen marketing claims of runtimes being the only one to support everything with no way of verifying that. publishing even bad tests helps give some transparency. It also encourages runtimes to have similar behavior which makes compiler discussions cheaper.

I know I will be uneasy when this extends to pre-release component things, for example testing ABIs which are not only incomplete, but also planned to change. However, I think that would be an acceptable downside to such a future site.

codefromthecrypt:
a meta point that might also be relevant, but easy to forget is this is a WebAssembly repo, not a bytecodealliance one. Ack that the people are the same who are administering it, but I also feel more comfortable about this being in the formal org than a third party one, both about the issue at hand: open to non-bytecodealliance requests including runtimes, as well requests to well not allow that ;) I suspect this also means that any future site will be on webassembly domain not bytecodealliance.

harry900831:
I'm the one who is adding this test suite to WasmEdge CI.

One major problem for me using WASI or implementing WASI API is lack of clear document and spec. Bring about the implementation for WASI function vary for each runtime and some common implementation mistakes that lots of runtime may encounter. For instance, we can see that lots of runtime has same bug about fd_readdir on large file number: https://github.com/bytecodealliance/wasmtime/issues/2493. 

It can make lots of things easier and prevent from past mistakes If we have an official WASI tests for runtime to verify their implementation. I believe this could benefit everyone and furthermore making WASI better.

To complete this goal, it would require discussion across different runtime or even WASI user's help to report an edge case that need to clarify while using WASI. Otherwise, this may end up a test suite that only serve wasmtime. It might be a good idea to attract others to contribute to this test suite by means of letting everyone know which runtime is participating.

In my opinion, making a website to let everyone know which runtime pass the test contains a sense of advertisement and running daily check seems to be a way to let others know which runtime is included in this project. Anyway, the main purpose is to involve major runtime in this test suite and complete it.

loganek:
Hi all, sorry for late reply.

Thanks a lot for all the input provided so far. I think for now we'll go ahead and accept different runtimes as part of daily job. There's apparently a lot of benefits, and I don't think this is a one-way-door decision. If at one point of time we think it's a better place for those daily runs (which I think it might be the case) or we shouldn't have those daily runs at all, I don't see any impact of removing them.

It's best interest for maintainers of runtimes to keep their adapters updated - if they are not, we'll consider temporarily or permanently deleting them to not confuse the users.

For now I'm going to merge this  PR.

loganek (approved):
@mathetake could you please rebase to latest main branch and fix conflicts?

mathetake:
done (I don't think there was conflict though)

codefromthecrypt:
@mathetake I think we should add our skips (check it out and then add during the run) even if we sometimes have skips empty. This would reduce reds around known issues. wdyt? https://github.com/tetratelabs/wazero/blob/main/.github/wasi_testsuite_skip.json

mathetake:
mind pushing the commit if you want? (I don't mind failing here or not temporarily)

codefromthecrypt:
sorry. I forgot this was on a shared branch. will do

codefromthecrypt:
actually good point, we can fail here. let's leave this one alone

loganek:
Yeah, ideally we shouldn't include any skips - if there are known issues in the runtime, I'd like them to be reflected in the result. If the fail is due to tests itself, we should rather fix the test.

loganek:
@mathetake I still can't merge it as it says
> This branch cannot be rebased due to conflicts

would you be able to rebase it again? I think we merged some PRs in the meantime. Sorry for inconvenience.

codefromthecrypt:
ps we pass all tests now (as of this hour anyway :))

mathetake:
ping @loganek can you merge?

loganek:
Hi @mathetake I still can't merge your PR due to unresolved conflicts. Possibly you're not rebasing to the right branch:
image

To speed things up, I've rebased your branch on top of main branch and opened a new PR (I've kept you as a commit author though): https://github.com/WebAssembly/wasi-testsuite/pull/71 Please review and if it looks ok, we'll merge that PR (and close this one).

mathetake:
ok cool thanks!

PR #56, opened 2023-02-20
doc: update dead link

loganek:


loganek:
Merging as it's trivial change and fixes the build.

PR #58, opened 2023-02-24
Update to wasi 0.11.

sunfishcode:
This ports bytecodealliance/wasmtime#5488 to the wasi-testsuite repo. There should be no significant behavior changes here, just slightly tidier APIs.


codefromthecrypt (approved):
indeed at least eyeballing this, there doesn't seem a behavior change

loganek (approved):

PR #67, opened 2023-03-20
Port `interesting_paths.rs` from Wasmtime's testsuite.

sunfishcode:
Following up on ac32f57400cdcdd0425d3085c24fc7fc40011d1c, this ports another test from Wasmtime's testsuite.


loganek (approved):

codefromthecrypt:
Is this missing a config file? When we run our runner it fails like this:

```
 Test interesting_paths failed
  [exit_code] 0 == 1
STDOUT:

STDERR:
usage: interesting_paths.wasm <scratch directory>
```

codefromthecrypt:
Thanks I just verified it works, though I have some work to do to get it to fail the intended way on windows

PR #68, opened 2023-03-20
Port more tests from Wasmtime's testsuite

sunfishcode:
Following up on ac32f57400cdcdd0425d3085c24fc7fc40011d1c, port several more tests from Wasmtime's testsuite to wasi-testsuite.

It is likely that at least some of these tests are testing behavior which differs between Wasm engine implementations. The intent here is not to instate these as authoritative, but to help surface these differences so that we can discuss whether it's best to change the test, or add documentation to the spec.

Other Wasm engines are welcome to submit their tests to wasi-testsuite as well.

### DIFF
--- a/tests/rust/src/bin/poll_oneoff_stdio.rs
+++ b/tests/rust/src/bin/poll_oneoff_stdio.rs
@@ -1,0 +1,137 @@
+use std::collections::HashMap;
+use std::mem::MaybeUninit;
+use wasi_tests::{assert_errno, STDERR_FD, STDIN_FD, STDOUT_FD};
+
+const TIMEOUT: u64 = 200_000_000u64; // 200 milliseconds, required to satisfy slow execution in CI
+const CLOCK_ID: wasi::Userdata = 0x0123_45678;
+const STDIN_ID: wasi::Userdata = 0x8765_43210;
+
+unsafe fn poll_oneoff_impl(r#in: &[wasi::Subscription]) -> Result<Vec<wasi::Event>, wasi::Errno> {
+    let mut out: Vec<wasi::Event> = Vec::new();
+    out.resize_with(r#in.len(), || {
+        MaybeUninit::<wasi::Event>::zeroed().assume_init()
+    });
+    let size = wasi::poll_oneoff(r#in.as_ptr(), out.as_mut_ptr(), r#in.len())?;
+    out.truncate(size);
+    Ok(out)
+}
+
+unsafe fn test_stdin_read() {
+    let clock = wasi::SubscriptionClock {
+        id: wasi::CLOCKID_MONOTONIC,
+        timeout: TIMEOUT,
+        precision: 0,
+        flags: 0,
+    };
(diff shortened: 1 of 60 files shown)


loganek:
@sunfishcode thanks for the PR. Do you already have a list of checks in the test that is not currently documented in the spec? I think that'd help with the discussion.

codefromthecrypt:
right now, tests don't run even on wasmtime on PR. can we please get this fixed?

Even if it is ok for some runtimes to fail, I think at least one should be able to pass, and without knowing that it is a hard to justify merging new tests.

codefromthecrypt:
Since the goal is discussion and acknowledged value in beyond wasmtime, I would suggest this:

1. these behaviors are not just about wasmtime, but they are also about rust and wasi-libc, other compiler owners may differ on the enforcements
2. if these are for discussion, it is better to split into different PRs vs a batch. Canonicalizing rules is an impactful thing, and each rule matters
3. ping diverse folks somehow (by CODEOWNERS or just ad-hoc). Don't leave it to watchers

There have been other changes that went in quickly today, they should have been looked at diversely before merge. Let's get a better practice in. It doesn't have to be as robust as for example Go's process, but it should have some diversity in it considering these rules define a large part of technology.

tests/rust/src/bin/poll_oneoff_stdio.rs
```diff
@@ -0,0 +1,137 @@
...
+    assert!(out.len() >= 1, "stdin read should return at least 1 event");
+    for event in out {
+        if event.type_ == wasi::EVENTTYPE_CLOCK {
+            assert_errno!(event.error, wasi::ERRNO_SUCCESS);
+            assert_eq!(
+                event.userdata, CLOCK_ID,
+                "the event.userdata should contain CLOCK_ID",
+            );
+        } else if event.type_ == wasi::EVENTTYPE_FD_READ {
+            assert_errno!(event.error, wasi::ERRNO_SUCCESS);
```
yamt, line 60: any idea about this? https://github.com/yamt/toywasm/blob/7e46dee9758f0f72a7d5f390790f35363b8feb5d/test/wasmtime-wasi-tests-blacklist.txt#L24-L30

yamt: FYI, i ended up with this: https://github.com/yamt/toywasm/commit/570e6706312cbdccb9d5268347186b441a24444b

sunfishcode:
> is it intentional that this batch doesn't include `poll_oneoff_files`? i'm asking because i have a portability concern about the test. cf. https://github.com/yamt/toywasm/blob/2ef7bfca8404de2cdf3e8e3faff3eb6011c4c5d2/test/wasmtime-wasi-tests-blacklist.txt#L15-L22

My idea here was just to port a lot of tests over to get things started, so I left out a few tests that I anticipated would be more involved.

loganek:
After reviewing those tests now I think it makes sense to have them merged. If there are any points for discussions, please open separate issue or PR.

PR #80, opened 2023-08-22
fix AssemblyScript build

evacchi:
I realized the build for AS tests is broken, this should fix it.

### DIFF
--- a/.github/workflows/compile-tests.yml
+++ b/.github/workflows/compile-tests.yml
@@ -24,6 +24,11 @@ jobs:
         with:
           node-version: 16
 
+      - name: Install dependencies
+        working-directory: tests/assemblyscript
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/compile-tests.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       - name: Check prettier codestyle
         working-directory: tests/assemblyscript
         run: |
+          npm install prettier
```
loganek, line 30: Thanks a lot for the change. I see prettier is already part of the package.json, so maybe we could run `npm install` command before calling `prettier-format-check` (e.g. have another step with `npm install`) instead of doing it in the `Build tests` step?

loganek (approved):
Many thanks!

Issue #42, opened 2023-01-11
timeout


yamt:
it would be nicer to have a timeout for each test cases.

vados-cosmonic:
We do have this now with the buckified test suite! Happy to re-open if it's still an issue, but tests do timeout right now.

PR #83, opened 2023-08-28
add a dumb timeout logic

yamt:
issue: https://github.com/WebAssembly/wasi-testsuite/issues/42

a few notes on this implementation:

* the timeout is hardcoded to 3 seconds for now. we can add a configuration in per-test json files when necessary.
* unit test doesn't cover the new code. i have no plan to write tests at this point.
* adapters may need something like https://github.com/yamt/toywasm/pull/117 for better cleanup

### DIFF
--- a/test-runner/wasi_test_runner/test_case.py
+++ b/test-runner/wasi_test_runner/test_case.py
@@ -24,6 +24,14 @@ class Result(NamedTuple):
         return len(self.failures) > 0
 
 
+class TimedoutResult(NamedTuple):
+    pass
+
+
+class SkippedResult(NamedTuple):
+    pass
+
+
 T = TypeVar("T", bound="Config")
 
 
--- a/test-runner/wasi_test_runner/test_case.py
+++ b/test-runner/wasi_test_runner/test_case.py
@@ -1,6 +1,6 @@
 import logging
 import json
-from typing import List, NamedTuple, TypeVar, Type, Dict, Any, Optional
+from typing import List, NamedTuple, TypeVar, Type, Dict, Any, Optional, Union
 
 
 class Output(NamedTuple):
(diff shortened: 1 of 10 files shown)


yamt:
@loganek please review when you have time. this is not urgent for me.

loganek:
Thanks for the change, and sorry for being late with the review. I'll try to get to it by end of this week.

yamt:
@loganek ping

test-runner/wasi_test_runner/test_case.py
```diff
@@ -63,5 +70,5 @@ def _validate_dict(cls: Type[T], dict_config: Dict[str, Any]) -> None:
 class TestCase(NamedTuple):
     name: str
     config: Config
-    result: Result
+    result: Union[Result | SkippedResult | TimedoutResult]
```
loganek, line 73: I've never seen this syntax before; IIRC it should be either `Union[Result, SkippedResult, TimedoutResult]` or `Result | SkippedResult | TimedoutResult`, see https://docs.python.org/3/library/typing.html#typing.Union

test-runner/wasi_test_runner/test_suite_runner.py
```diff
@@ -77,13 +80,17 @@ def _execute_single_test(
 ) -> TestCase:
     config = _read_test_config(test_path)
     test_start = time.time()
-    test_output = runtime.run_test(test_path, config.args, config.env, config.dirs)
+    try:
+        test_output = runtime.run_test(test_path, config.args, config.env, config.dirs)
+        result: Union[Result | TimedoutResult] = _validate(validators, config, test_output)
```
loganek, line 85: same here: https://github.com/WebAssembly/wasi-testsuite/pull/83/files#r1471960979

vados-cosmonic:
Closing this out as the test suite has changed dramatically (been buckified, scripts work differently, etc), and also it's been 3 years.

PR #85, opened 2023-09-11
Fix tests on WAMR

zoraaver:
Most of the changes are adding rights to fix permissions issues and changing expected error codes.

I've also changed the `stdio.rs` test so that the destination (`to`) file descriptors are valid. The test was previously failing on WAMR since the destination fd's were arbitrary numbers which seems to be explicitly not permitted by the [docs](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/witx/wasi_snapshot_preview1.witx#L245-L251). A test case has been added in `renumber.rs` to enforce that renumbering to invalid destination file descriptors is an error.

If we want to align on return error codes between the runtimes for certain filesystem functions, I don't think it would be too much work to change the WAMR implementation to make it fall in line with wasmtime but for the moment have raised this PR to get some clarification and feedback.

### DIFF
--- a/tests/rust/src/bin/interesting_paths.rs
+++ b/tests/rust/src/bin/interesting_paths.rs
@@ -16,7 +16,8 @@ unsafe fn test_interesting_paths(dir_fd: wasi::Fd, arg: &str) {
     assert_errno!(
         wasi::path_open(dir_fd, 0, "/dir/nested/file", 0, 0, 0, 0)
             .expect_err("opening a file with an absolute path"),
-        wasi::ERRNO_PERM
+        wasi::ERRNO_PERM,
+        wasi::ERRNO_NOTCAPABLE
     );
 
     // Now open it with a path containing "..".
@@ -83,7 +84,8 @@ unsafe fn test_interesting_paths(dir_fd: wasi::Fd, arg: &str) {
     assert_errno!(
         wasi::path_open(dir_fd, 0, &bad_path, 0, 0, 0, 0)
             .expect_err("opening a file with too many \"..\"s in the path should fail"),
-        wasi::ERRNO_PERM
+        wasi::ERRNO_PERM,
+        wasi::ERRNO_NOTCAPABLE
     );
     wasi::path_unlink_file(dir_fd, "dir/nested/file")
         .expect("unlink_file on a symlink should succeed");
--- a/tests/rust/src/bin/readlink.rs
+++ b/tests/rust/src/bin/readlink.rs
@@ -1,5 +1,5 @@
 use std::{env, process};
-use wasi_tests::{assert_errno, create_file, create_tmp_dir, open_scratch_directory};
+use wasi_tests::{create_file, create_tmp_dir, open_scratch_directory};
 
 unsafe fn test_readlink(dir_fd: wasi::Fd) {
     // Create a file in the scratch directory.
@@ -20,12 +20,14 @@ unsafe fn test_readlink(dir_fd: wasi::Fd) {
         "the remaining bytes should be untouched"
     );
 
-    // Read link into smaller buffer than the actual link's length
+    // Read link into smaller buffer than the actual link's length. The first
+    // buf_len bytes of the link content should be placed in the buffer
     let buf = &mut [0u8; 4];
-    let err = wasi::path_readlink(dir_fd, "symlink", buf.as_mut_ptr(), buf.len())
-        .err()
-        .expect("readlink with too-small buffer should fail");
-    assert_errno!(err, wasi::ERRNO_RANGE);
+    let bufused = wasi::path_readlink(dir_fd, "symlink", buf.as_mut_ptr(), buf.len())
+        .expect("readlink with buffer smaller than link content failed");
+
+    assert_eq!(bufused, 4, "should use 4 bytes of the buffer");
+    assert_eq!(&buf[..4], b"targ", "buffer should contain 'targ'");
 
     // Clean up.
     wasi::path_unlink_file(dir_fd, "target").expect("removing a file");
--- a/tests/rust/src/bin/stdio.rs
+++ b/tests/rust/src/bin/stdio.rs
@@ -1,13 +1,49 @@
-use wasi_tests::{STDERR_FD, STDIN_FD, STDOUT_FD};
+use std::{env, process};
 
-unsafe fn test_stdio() {
-    for fd in &[STDIN_FD, STDOUT_FD, STDERR_FD] {
-        wasi::fd_fdstat_get(*fd).expect("fd_fdstat_get on stdio");
-        wasi::fd_renumber(*fd, *fd + 100).expect("renumbering stdio");
+use wasi_tests::{create_tmp_dir, open_scratch_directory, STDERR_FD, STDIN_FD, STDOUT_FD};
+
+const TEST_FILENAME: &'static str = "file.cleanup";
+
+unsafe fn test_stdio(dir_fd: wasi::Fd) {
+    for stdio_from_fd in &[STDIN_FD, STDOUT_FD, STDERR_FD] {
+        let stdio_to_fd = wasi::path_open(dir_fd, 0, TEST_FILENAME, wasi::OFLAGS_CREAT, 0, 0, 0)
+            .expect("open file");
+
+        wasi::fd_renumber(*stdio_from_fd, stdio_to_fd).expect("renumbering stdio");
+
+        wasi::fd_fdstat_get(stdio_to_fd).expect("fd_fdstat_get failed");
+        wasi::fd_fdstat_get(*stdio_from_fd).expect_err("stdio_from_fd is not closed");
+
+        // Cleanup
+        wasi::path_unlink_file(dir_fd, "file").expect("failed to remove file");
     }
 }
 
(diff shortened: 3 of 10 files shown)


tests/rust/src/bin/readlink.rs
```diff
@@ -20,13 +20,6 @@ unsafe fn test_readlink(dir_fd: wasi::Fd) {
         "the remaining bytes should be untouched"
     );
 
-    // Read link into smaller buffer than the actual link's length
-    let buf = &mut [0u8; 4];
-    let err = wasi::path_readlink(dir_fd, "symlink", buf.as_mut_ptr(), buf.len())
-        .err()
-        .expect("readlink with too-small buffer should fail");
-    assert_errno!(err, wasi::ERRNO_RANGE);
```
zoraaver, line 28: On [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/functions/readlink.html) and the current WAMR implementation, this doesn't cause an error - the buffer is filled with first `buflen` bytes from the link.

tests/rust/src/bin/stdio.rs
```diff
@@ -1,13 +0,0 @@
-use wasi_tests::{STDERR_FD, STDIN_FD, STDOUT_FD};
-
-unsafe fn test_stdio() {
-    for fd in &[STDIN_FD, STDOUT_FD, STDERR_FD] {
-        wasi::fd_fdstat_get(*fd).expect("fd_fdstat_get on stdio");
-        wasi::fd_renumber(*fd, *fd + 100).expect("renumbering stdio");
```
loganek, line 6: This test looks fine to me; it validates whether `fd_renumber` works on STDIO file descriptors (we have another `renumber.rs` test that operates on file descriptors). As of today, nothing in the documentation says this function should not work on those descriptors; I'd suggest first opening a PR that clarifies the behavior here: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/witx/wasi_snapshot_preview1.witx if you believe this function should not work on STDIO; unless it's specified, I'd treat stdio as regular file descriptors here (i.e. the test should work).

zoraaver: Sorry, I should have been more clear - the issue isn't with the STDIO file descriptors. It's that the destination file descriptor is an arbitrary number i.e. it doesn't refer to any actual open file descriptor.

The [docs](https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/witx/wasi_snapshot_preview1.witx#L245-L251) imply that this behaviour is not supported if I understand correctly.

Although actually, if we still want to validate that renumbering stdio fd's is possible, I can alter the test to renumber the stdio fd's to valid fd's.

zoraaver: I've changed the test so that the destination file descriptors are valid.

tests/rust/src/bin/readlink.rs
loganek: Could we open a PR for the doc to clarify that then?

tests/rust/src/bin/stdio.rs
```diff
@@ -1,13 +1,47 @@
...
+use wasi_tests::{create_tmp_dir, open_scratch_directory, STDERR_FD, STDIN_FD, STDOUT_FD};
+
+unsafe fn test_stdio(dir_fd: wasi::Fd) {
+    for stdio_from_fd in &[STDIN_FD, STDOUT_FD, STDERR_FD] {
+        let stdio_to_fd =
+            wasi::path_open(dir_fd, 0, "file", wasi::OFLAGS_CREAT, 0, 0, 0).expect("open file");
```
loganek, line 8: Any temporary files should have a `.cleanup` suffix. See https://github.com/WebAssembly/wasi-testsuite/blob/main/doc/specification.md

tests/rust/src/bin/stdio.rs
```diff
@@ -1,13 +1,47 @@
...
+
+        wasi::fd_renumber(*stdio_from_fd, stdio_to_fd).expect("renumbering stdio");
+
+        wasi::fd_fdstat_get(stdio_to_fd).expect("fd_fdstat_get failed");
+        wasi::fd_fdstat_get(*stdio_from_fd).expect_err("stdio_from_fd is not closed");
```
loganek, line 13: Is it expected that the source fd is being closed in that case?

tests/rust/src/lib.rs
```diff
@@ -39,6 +39,12 @@ pub unsafe fn create_tmp_dir(dir_fd: wasi::Fd, name: &str) -> wasi::Fd {
             | wasi::RIGHTS_FD_READDIR
             | wasi::RIGHTS_FD_FILESTAT_GET
             | wasi::RIGHTS_FD_SEEK
+            | wasi::RIGHTS_PATH_LINK_SOURCE
```
loganek, line 42: Given that those tests were passing on  wasmtime, I wonder if wasmtime correctly implemented file right system. Looks like files created in parent directory inherit `base` rights, instead of inheriting `inheriting` rights. @abrown is that something you could confirm?

tests/rust/src/bin/interesting_paths.rs
```diff
@@ -16,7 +16,8 @@ unsafe fn test_interesting_paths(dir_fd: wasi::Fd, arg: &str) {
     assert_errno!(
         wasi::path_open(dir_fd, 0, "/dir/nested/file", 0, 0, 0, 0)
             .expect_err("opening a file with an absolute path"),
-        wasi::ERRNO_PERM
+        wasi::ERRNO_PERM,
+        wasi::ERRNO_NOTCAPABLE
```
loganek, line 20: I think it's probably fine to have both for now, but I'd recommend opening an issue in the [WASI](https://github.com/WebAssembly/WASI/) repo to clarify that. Ideally this should be consistent.

tests/rust/src/bin/readlink.rs
zoraaver: Sure, https://github.com/WebAssembly/WASI/pull/554. I've also amended the test to enforce the POSIX/WAMR behaviour since it's more in keeping with docs; I would have thought that if we mention a POSIX function, it's safe to assume the behaviour is the same _unless_ specified otherwise.

tests/rust/src/bin/stdio.rs
zoraaver: Yes, that's the behaviour that's currently enforced by the [renumber](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/src/bin/renumber.rs#L49-L52) test. The docs don't explicitly mention closing the source fd, but `fd_renumber` is meant to be analogous (with a few differences) to [dup2](https://pubs.opengroup.org/onlinepubs/009604499/functions/dup.html), which states that the source fd will be closed.

tests/rust/src/bin/interesting_paths.rs
zoraaver: https://github.com/WebAssembly/WASI/issues/555

tests/rust/src/bin/stdio.rs
zoraaver: Done

tests/rust/src/lib.rs
abrown: Can you point me to which version of Wasmtime is being used to run the tests? I do see recent-ish PRs that may have changed this behavior:
- https://github.com/bytecodealliance/wasmtime/pull/6784
- https://github.com/bytecodealliance/wasmtime/pull/6479

tests/rust/src/bin/stdio.rs
loganek: I wonder then whether we should also have a test where we renumber to an arbitrary file descriptor and in that case expect the failure?

zoraaver: Yes, that's correct - we should expect failure if the destination is an invalid fd. I'll add a test case in the existing `renumber.rs` test.

tests/rust/src/lib.rs
loganek: This is on 12.0.2

loganek (approved):
LGTM, thanks

tests/rust/src/lib.rs
abrown: I would try the latest version of Wasmtime then and see if that resolves things?

loganek: I just tried the latest main branch, and tests in its current form (e.g. `path_filestat`) are still passing.

zoraaver: > Looks like files created in parent directory inherit base rights, instead of inheriting inheriting rights.

Just want to confirm that's the behaviour I observed.

@abrown if it's helpful, here's a list of tests which should fail (without the changes from this PR) and the required missing rights:

* `path_link.rs`
```
wasi::RIGHTS_PATH_LINK_SOURCE
wasi::RIGHTS_PATH_LINK_TARGET
wasi::RIGHTS_PATH_OPEN
wasi::RIGHTS_PATH_UNLINK_FILE
```
* `path_filestat` and `symlink_filestat.rs`
```
wasi::RIGHTS_PATH_FILESTAT_GET
```

loganek (approved):

PR #86, opened 2023-09-12
Port some more wasmtime filesystem tests

zoraaver:
To close the testing gaps on any filesystem functions, port some more of the wasmtime tests. A few test cases in dir_fd_op_failures seemed to be specific to wasmtime; namely fd_sync, fd_datasync, fd_fdstat_set_flags and fd_advise. These functions return errors when called on a directory fd on wasmtime but on WAMR (and POSIX) they return 0 so for the moment have removed them.

### DIFF
--- a/tests/rust/src/bin/dir_fd_op_failures.rs
+++ b/tests/rust/src/bin/dir_fd_op_failures.rs
@@ -1,0 +1,128 @@
+use std::{env, process};
+use wasi_tests::{assert_errno, open_scratch_directory};
+
+unsafe fn test_fd_dir_ops(dir_fd: wasi::Fd) {
+    let stat = wasi::fd_filestat_get(dir_fd).expect("failed to fdstat");
+    assert_eq!(stat.filetype, wasi::FILETYPE_DIRECTORY);
+
+    let (pr_fd, pr_name_len) = (3..)
+        .map_while(|fd| wasi::fd_prestat_get(fd).ok().map(|stat| (fd, stat)))
+        .find_map(|(fd, wasi::Prestat { tag, u })| {
+            (tag == wasi::PREOPENTYPE_DIR.raw()).then_some((fd, u.dir.pr_name_len))
+        })
+        .expect("failed to find preopen directory");
+
+    let mut pr_name = vec![];
+    let r = wasi::fd_prestat_dir_name(pr_fd, pr_name.as_mut_ptr(), 0);
+    assert_eq!(r, Err(wasi::ERRNO_NAMETOOLONG));
+
+    // Test that passing a larger than necessary buffer works correctly
+    let mut pr_name = vec![0; pr_name_len + 1];
+    let r = wasi::fd_prestat_dir_name(pr_fd, pr_name.as_mut_ptr(), pr_name_len + 1);
+    assert_eq!(r, Ok(()));
+
+    let mut read_buf = vec![0; 128].into_boxed_slice();
+    let iovec = wasi::Iovec {
+        buf: read_buf.as_mut_ptr(),
+        buf_len: read_buf.len(),
+    };
+
+    // On posix, this fails with ERRNO_ISDIR
+    assert_errno!(
+        wasi::fd_read(dir_fd, &[iovec]).expect_err("fd_read error"),
+        wasi::ERRNO_BADF,
+        wasi::ERRNO_NOTCAPABLE
+    );
+
+    // On posix, this fails with ERRNO_ISDIR
+    assert_errno!(
+        wasi::fd_pread(dir_fd, &[iovec], 0).expect_err("fd_pread error"),
+        wasi::ERRNO_BADF,
+        wasi::ERRNO_NOTCAPABLE
+    );
+
+    let write_buf = vec![0; 128].into_boxed_slice();
+    let ciovec = wasi::Ciovec {
+        buf: write_buf.as_ptr(),
+        buf_len: write_buf.len(),
+    };
+
+    assert_errno!(
+        wasi::fd_write(dir_fd, &[ciovec]).expect_err("fd_write error"),
+        wasi::ERRNO_BADF,
+        wasi::ERRNO_NOTCAPABLE
+    );
+
+    assert_errno!(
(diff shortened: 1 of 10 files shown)


tests/rust/src/bin/dir_fd_op_failures.rs
```diff
@@ -0,0 +1,128 @@
...
+    let mut read_buf = vec![0; 128].into_boxed_slice();
+    let iovec = wasi::Iovec {
+        buf: read_buf.as_mut_ptr(),
+        buf_len: read_buf.len(),
+    };
+
+    // On posix, this fails with ERRNO_ISDIR
+    assert_errno!(
+        wasi::fd_read(dir_fd, &[iovec]).expect_err("fd_read error"),
+        wasi::ERRNO_BADF,
```
loganek, line 33: as mentioned in the other PR, we probably want a WASI issue to clarify this in documentation.

loganek (approved):
Please review all the temporary files/directories and use .cleanup suffix (see https://github.com/WebAssembly/wasi-testsuite/blob/main/doc/specification.md)

tests/rust/src/bin/remove_directory_trailing_slashes.rs
```diff
@@ -0,0 +1,60 @@
+use std::{env, process};
+use wasi_tests::{assert_errno, create_file, open_scratch_directory};
+
+unsafe fn test_remove_directory_trailing_slashes(dir_fd: wasi::Fd) {
+    // Create a directory in the scratch directory.
+    wasi::path_create_directory(dir_fd, "dir").expect("creating a directory");
```
loganek, line 6: Please use `.cleanup` suffix for temporary files and directories.

tests/rust/src/bin/remove_directory_trailing_slashes.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    // Test that removing it succeeds.
+    wasi::path_remove_directory(dir_fd, "dir")
```
loganek, line 9: Please use `.cleanup` suffix for temporary files and directories.

tests/rust/src/bin/remove_directory_trailing_slashes.rs
```diff
@@ -0,0 +1,60 @@
...
+        .expect("remove_directory on a directory should succeed");
+
+    wasi::path_create_directory(dir_fd, "dir").expect("creating a directory");
```
loganek, line 12: Please use `.cleanup` suffix for temporary files and directories.

tests/rust/src/bin/remove_directory_trailing_slashes.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    // Test that removing it with a trailing slash succeeds.
+    wasi::path_remove_directory(dir_fd, "dir/")
+        .expect("remove_directory with a trailing slash on a directory should succeed");
+
+    // Create a temporary file.
+    create_file(dir_fd, "file");
```
loganek, line 19: Please use `.cleanup` suffix for temporary files and directories.

tests/rust/src/bin/path_open_read_write.rs
```diff
@@ -0,0 +1,139 @@
+use std::{env, process};
+use wasi_tests::{assert_errno, create_file, open_scratch_directory};
+
+unsafe fn test_path_open_read_write(dir_fd: wasi::Fd) {
+    create_file(dir_fd, "file");
```
loganek, line 5: Please use `.cleanup` suffix for temporary files and directories.

tests/rust/src/bin/dir_fd_op_failures.rs
zoraaver: Opened https://github.com/WebAssembly/WASI/issues/555

tests/rust/src/bin/path_open_preopen.rs
```diff
@@ -0,0 +1,153 @@
+const FIRST_PREOPEN: u32 = 3;
+
+unsafe fn path_open_preopen() {
+    let prestat = wasi::fd_prestat_get(FIRST_PREOPEN).expect("fd 3 is a preopen");
```
loganek, line 4: WASI defines file descriptors 0, 1, and 2 for stdio, but it doesn't say that further preopens will have a sequence number: https://github.com/WebAssembly/WASI/blob/60cff3164f66d15092248a5adfa901a4caa7fea8/legacy/application-abi.md?plain=1#L54

`path_open`, even though it's not quite the same, explicitly specifies that the number will be randomized: https://github.com/WebAssembly/WASI/blob/main/legacy/preview1/witx/wasi_snapshot_preview1.witx#L355

zoraaver: Good point - I've added a function which runs through all possible values of an `fd` starting from 3 and returns the first one which is a preopen. 

I think this is fine for the moment since wasmtime and WAMR seem to open preopened fd's in sequence after the stdio ones anyway. But it might be an issue if another runtime really did choose a random number since the function would take a long time to find the first preopened fd 😄 given there are 2^31 possibilities.

tests/rust/src/bin/path_open_preopen.rs
```diff
@@ -0,0 +1,171 @@
+unsafe fn find_first_preopened_fd() -> Result<wasi::Fd, String> {
```
loganek, line 1: I think this looks more secure; however, we should probably update the [documentation](https://github.com/WebAssembly/wasi-testsuite/blob/main/doc/specification.md) to say that preopens defined in `dir` should be **the only** preopens in the runtime; runtime can choose to e.g. by default preopen current directory since the WASI spec doesn't specify that AFAIK. Alternatively, you can also pass the dir name as a CLI argument and use `fd_prestat_dir_name` to make sure the preopen found is the right one.

zoraaver: Thanks, that makes sense. I'll go with the second option of checking the directory name since that seems to be what `open_scratch_directory` (used by most of the tests) does.

zoraaver: Changed so that it matches based on dir name from the CLI args 👍

loganek (approved):
LGTM

PR #87, opened 2023-09-12
Add more test cases for fd rights

zoraaver:
Remove the previous path_open_read_without_rights.rs test and replace it with a similar test rights.rs which tests all cases from the previous test and some more cases. It verifies that reading/writing with sufficient rights functions correctly and that dropping rights on an fd forbids operations on the fd which would have been previously allowed.


loganek (approved):
LGTM

PR #91, opened 2023-09-22
revert local override on no_fdflags_sync_support

yamt:
* I couldn't find this in wasmtime repo. I suppose this was a local modification done as a part of https://github.com/WebAssembly/wasi-testsuite/pull/68

* As some implementations have implemented it, it isn't appropriate to assert NOTSUP by default.


loganek:
I think this change is OK given that this is specifically for preview1 (because of the use of wasi crate). 

I think we should also somehow keep track on the test configuration in test results, I've opened a ticket to discuss this: https://github.com/WebAssembly/wasi-testsuite/issues/93

zoraaver:
I'm just wondering whether we should assert the sync flag behaviour at all. The test which uses this is `path_filestat.rs`. Since the tests can't actually assert whether the sync flags work as intended, maybe we should remove that test instead? It has limited value and support is dependent on the host platform.

PR #92, opened 2023-09-22
Add expected error code for fd_prestat_dir_name

zoraaver:
In the case that the user provides a buffer which is too small to fit the directory name, it makes more sense to return EINVAL rather than ENAMETOOLONG. ENAMETOOLONG has the speficic connotation that a path was longer than the maximum size permitted by the OS - that doesn't really match this error condition. We're keeping both error codes for now since we don't specify in the docs what error code should be returned.

### DIFF
--- a/tests/rust/src/bin/dir_fd_op_failures.rs
+++ b/tests/rust/src/bin/dir_fd_op_failures.rs
@@ -13,8 +13,12 @@ unsafe fn test_fd_dir_ops(dir_fd: wasi::Fd) {
         .expect("failed to find preopen directory");
 
     let mut pr_name = vec![];
-    let r = wasi::fd_prestat_dir_name(pr_fd, pr_name.as_mut_ptr(), 0);
-    assert_eq!(r, Err(wasi::ERRNO_NAMETOOLONG));
+    assert_errno!(
+        wasi::fd_prestat_dir_name(pr_fd, pr_name.as_mut_ptr(), 0)
+            .expect_err("fd_prestat_dir_name error"),
+        wasi::ERRNO_INVAL,
+        wasi::ERRNO_NAMETOOLONG
+    );
 
     // Test that passing a larger than necessary buffer works correctly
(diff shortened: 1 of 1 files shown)


loganek (approved):
LGTM

yamt:
is it strictly mentioned in the spec?
otherwise, i guess it's safer to allow both error codes.
wasmtime returns ENAMETOOLONG, doesn't it?

zoraaver:
It's not mentioned in the spec, https://docs.rs/wasi/latest/wasi/fn.fd_prestat_dir_name.html. But I'd say it's better to standardise on a single error code. In this case it shouldn't be much work to change it either way. As far as I understand, the purpose of the tests isn't so that they pass immediately on all runtimes, it's to highlight and standardise any differences in behaviour. 

> wasmtime returns ENAMETOOLONG, doesn't it?

It probably does since this test is originally from the wasmtime set of tests.

I've raised https://github.com/WebAssembly/WASI/issues/555 to discuss standardising error codes from filesystem functions in general.

yamt:
> It's not mentioned in the spec, https://docs.rs/wasi/latest/wasi/fn.fd_prestat_dir_name.html. But I'd say it's better to standardise on a single error code. In this case it shouldn't be much work to change it either way. As far as I understand, the purpose of the tests isn't so that they pass immediately on all runtimes, it's to highlight and standardise any differences in behaviour.

my point is such a standardization should happen in spec, not tests.

> 
> > wasmtime returns ENAMETOOLONG, doesn't it?
> 
> It probably does since this test is originally from the wasmtime set of tests.
> 
> I've raised [WebAssembly/WASI#555](https://github.com/WebAssembly/WASI/issues/555) to discuss standardising error codes from filesystem functions in general.

i see.

zoraaver:
> my point is such a standardization should happen in spec, not tests.

True, I've kept both error codes until we decide how to proceed with standardising error codes in general.

PR #94, opened 2023-11-20
Add tests for sock_shutdown() API

loganek:
According to the specification [1] the API should behave similarly to the POSIX one, therefore expecting specific errno codes

https://github.com/WebAssembly/WASI/blob/2980bb39e1d2a4a2adae4748908cb4325cd41a26/legacy/preview1/docs.md#sock_shutdown

### DIFF
--- a/tests/c/testsuite/sock_shutdown-not_sock.c
+++ b/tests/c/testsuite/sock_shutdown-not_sock.c
@@ -1,0 +1,12 @@
+#include <assert.h>
+#include <errno.h>
+#include <stdlib.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
+int main() {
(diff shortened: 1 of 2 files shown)


tests/c/testsuite/sock_shutdown-not_sock.c
```diff
@@ -0,0 +1,14 @@
+#include <assert.h>
+#include <errno.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/socket.h>
+
+int main() {
+  int fd = fileno(stdout);
```
yamt, line 8: it's simpler to use STDOUT_FILENO

yamt (approved):
looks reasonable to me

PR #95, opened 2023-12-04
Add test for pwrite with append flag

yagehu:
WASI specifies `fd_pwrite` to be similar to [POSIX `pwritev`][1] which specifies that it should ignore the append flag and write from the supplied offset.  Currently, only Wasmtime implements this behavior.

NB: There is a documented [Linux bug][2] that pwrite does not respect the offset and instead appends.  Wasmtime fixes this by tracking the append flag status and hiding it from the OS.

[1]: https://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html
[2]: https://linux.die.net/man/2/pwrite


yagehu:
I've filed an issue with the WasmEdge team. They [have expressed](https://github.com/WasmEdge/WasmEdge/issues/3062) desire to bring WasmEdge behavior in line with Wasmtime and add this test case.

loganek:
Thanks, the test looks reasonable. Could you please fix formatting problems reported by the build?

yagehu:
@loganek Thanks! Fixed format.

PR #96, opened 2023-12-09
Assert error on ambiguous fstflags

yagehu:
This commit adds a Rust test case that asserts an ambiguous `fstflags` input to `fd_filestat_set_times` should result in `inval` errno. An `fstflags` argument is ambiguous if:

1. both `atim` and `atim_now` are set, or
2. both `mtim` and `mtim_now` are set.

This assertion is consistent with Wasmtime and WasmEdge's current behavior.


loganek (approved):
LGTM

yagehu:
Can anyone merge this?

loganek:
Sorry for missing that, merged

PR #98, opened 2024-01-08
dir_fd_op_failures.rs: use non-zero len to match the comment above

yamt:
Note: on posix, len=0 may fail with EINVAL.


yamt:
@loganek ping

loganek (approved):
LGTM

PR #99, opened 2024-02-27
Add support for test config overrides

zoraaver:
It can be useful to override the config for certain tests. The main use case for this is to override environment variables which control test assertions (e.g. `NO_DANGLING_FILESYSTEM`, `ERRNO_MODE_WINDOWS`). When running the tests in CI, it would be nicer to support this without manually modifying the JSON test config files.

Any feedback would be appreciated, thanks.

### DIFF
--- a/test-runner/wasi_test_runner/test_suite_runner.py
+++ b/test-runner/wasi_test_runner/test_suite_runner.py
@@ -6,7 +6,7 @@ import shutil
 import time
 
 from datetime import datetime
-from typing import List, cast
+from typing import List, Optional, cast
 
 from .filters import TestFilter
 from .runtime_adapter import RuntimeAdapter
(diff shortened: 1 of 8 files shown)


test-runner/wasi_test_runner/test_suite_runner.py
```diff
@@ -73,9 +81,12 @@ def _skip_single_test(
 
 
 def _execute_single_test(
-    runtime: RuntimeAdapter, validators: List[Validator], test_path: str
+    runtime: RuntimeAdapter,
+    validators: List[Validator],
+    test_path: str,
+    config_override: Optional[Config],
```
loganek, line 87: given that this eventually is optional anyway, could we perhaps remove the `StubConfigOverride` class and pass around `None` if no override is provided?

vados-cosmonic:
The new refactored test suite does allow for test (and runner-level) overrides! It's been 2 years but i think the need for this is as good as solved.

PR #103, opened 2025-04-05
pwrite-with-append.c: ensure that O_APPEND is actually working

yamt:
the old code would pass for an implementation which ignores O_APPEND.


loganek (approved):
LGTM

PR #104, opened 2025-04-05
ci: bump the version of a few actions

yamt:
cf. https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

also, update rust wasi target name.


loganek (approved):
LGTM

PR #110, opened 2025-08-14
Update to wasi-sdk 27, for test compilation

wingo:
Also, pass --target=wasm32-wasip1 to build.sh, given that the existing tests target that version.


wingo:
My local tests using a fresh-built wasi-sdk exhibit a couple failures (e.g. related to `ENOTCAPABLE`, which wasi-libc doesn't use any more).  Let's see what happens.

pchickey (approved):

PR #112, opened 2025-08-15
Build C test cases via Python script to target-specific dir

wingo:
We move all the C sources from tests/c/testsuite to tests/c/src, somewhat like the Rust tests, then treat tests/c/testsuite as a built directory.

The build script is rewritten to Python, as it is getting a bit more complicated.

Tests are compiled into target-specific dirs, currently only tests/c/testsuite/wasm32-wasip1.

The build script writes the version-specific manifest.json, so as to include the target in the test suite name.

Workflows updated appropriately.

Stacked on https://github.com/WebAssembly/wasi-testsuite/pull/110.


tests/c/src/fopen-with-no-access.c
```diff
@@ -7,7 +7,7 @@ int main() {
   FILE *file = fopen("fs-tests.dir/file", "r");
 
   assert(file == NULL);
-  assert(errno == ENOTCAPABLE);
+  assert(errno == ENOTCAPABLE || errno == ENOENT);
```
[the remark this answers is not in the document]
pchickey: Yes, in wasip2 and beyond we got rid of the rights system from p1, so ENOTCAPABLE is no longer an applicable errno. It shouldn't be used at all in wasip2 and beyond.

tests/c/build.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env python3
```
loganek, line 1: just looking at the https://github.com/WebAssembly/wasi-testsuite/pull/114/files and it seems a lot of duplication, I wonder if we can refactor it in a way that the common part can be reused across different languages?

wingo: I think yes, but wdyt about me doing that in a followup?  It's a bit easier when the three implementations are actually there in the source tree.

.gitignore
```diff
@@ -1,2 +1,3 @@
 *.wasm
-.DS_Store
\ No newline at end of file
+.DS_Store
+/tests/c/testsuite/
```
loganek, line 3: perhaps it's best to use relative path here?

tests/c/src/fopen-with-no-access.c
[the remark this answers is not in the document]
loganek: should this be reverted then given it's weakening the assertion for p1?

PR #113, opened 2025-08-20
Use a higher fd in sock_shutdown-invalid_fd.c

Virv12:
By my understanding fds starting from 3 refer to preopened directories.
Thus accessing fd 3 will cause an ENOTSOCK and fail the test.
Using a higher fd solves this issue.

If there's any problem with this PR please let me know.


Virv12:
> Do you have a specific example of a command where this particular test fails?

Not really, I'm writing my own wasm executor, but I don't think it counts.

> For this test, from what I can see, we don't request preopening any particular directories.

I guess I should fix my implementation to follow this.

PR #114, opened 2025-08-21
Per-target rust testsuite

wingo:
As in https://github.com/WebAssembly/wasi-testsuite/pull/112, we reorganize the directory structure and change to a python build script, with the goal of making space for wasip3.

Before, it was:
  - tests/rust/: Root of cargo package

  - tests/rust/src: Rust source files

  - tests/rust/testsuite: JSON test files, fs-tests.dir; built .wasm files dumped here

Now it is:
  - tests/rust/wasm32-wasip1: Root of cargo package for wasip1 builds

  - tests/rust/wasm32-wasip1/src: Rust source, but also JSON and fs-tests.dir

  - tests/rust/testsuite/: Removed; instead it is a built dir

  - tests/rust/testsuite/wasm32-wasip1: All wasm and json files copied here

Note, the workflows are not yet updated.


tests/rust/build.py
```diff
@@ -0,0 +1,94 @@
...
+SYSTEMS = ['wasm32']
+VERSIONS = ['wasip1'] # + ['wasip2', 'wasip3']
+
+def compute_target(system, version):
+    return f"{system}-{version}"
+
+def compute_build_target(system, version):
+    if version == 'wasip3':
+        # wasm32-wasip3 triple not yet supported.
+        return compute_target(system, 'wasip2')
```
loganek, line 27: if wasm32-wasip3 is not supported, then we should throw an error, instead of using another version. But as I stated in my previous comment, we could perhaps remove this `if` statement completely.

tests/rust/build.py
```diff
@@ -0,0 +1,94 @@
...
+import shlex
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--dry-run", action="store_true")
+parser.add_argument("--verbose", action="store_true")
+parser.add_argument("--release", action="store_true")
```
loganek, line 14: Consider adding `help` for each of those, not all the parameters are self-explanatory

tests/rust/build.py
```diff
@@ -0,0 +1,94 @@
...
+def compute_target(system, version):
+    return f"{system}-{version}"
+
+def compute_build_target(system, version):
+    if version == 'wasip3':
+        # wasm32-wasip3 triple not yet supported.
+        return compute_target(system, 'wasip2')
+    return compute_target(system, version)
+
+BASE_DIR = Path(__file__).parent
```
loganek, line 30: That could be a cli argument with the default value set to `Path(__file__).parent` for local executions.

tests/rust/build.py
```diff
@@ -0,0 +1,94 @@
...
+
+def run(argv):
+    if args.verbose:
+        print(shlex.join([str(x) for x in argv]))
+    if not args.dry_run:
+        r = subprocess.run(argv)
+        if r.returncode != 0:
+            sys.exit(r.returncode)
```
loganek, line 38: This should have similar effect
```suggestion
        r = subprocess.run(argv, check=True)
```

tests/rust/build.py
```diff
@@ -0,0 +1,94 @@
...
+args = parser.parse_args()
+
+SYSTEMS = ['wasm32']
+VERSIONS = ['wasip1'] # + ['wasip2', 'wasip3']
+
+def compute_target(system, version):
+    return f"{system}-{version}"
+
+def compute_build_target(system, version):
+    if version == 'wasip3':
```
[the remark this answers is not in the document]
wingo: I'm going to add calls for it soon.  The goal is to test wasip3, I need a dir for wasip3 tests, but I have to pass `--target=wasm32-wasip2` to the compiler until 0.3.0 is blessed.  Trust me on this one :)

tests/rust/build.py
wingo: See above.

tests/rust/build.py
wingo: Hmm, I kept the exit so as not to show the user a traceback, but I added a print for the command-line that could be re-run.

wingo:
Rebased to resolve conflict, no other change

loganek (approved):

PR #115, opened 2025-08-21
Per-target Assemblyscript testsuite

wingo:
As in https://github.com/WebAssembly/wasi-testsuite/pull/112, we reorganize the directory structure and change to a python build script, with the goal of harmonizing the various test suites, so that testsuite/ is always a built dir.

Before, it was:
  - tests/assemblyscript/: Root of npm repo

  - tests/assemblyscript/testsuite: AS source files, JSON test files; built .wasm files dumped here

Now it is:
  - tests/assemblyscript/wasm32-wasip1: Root of npm package for wasip1 builds

  - tests/assemblyscript/wasm32-wasip1/src: Assemblyscript source, but also JSON

  - tests/assemblyscript/testsuite/: Removed; instead it is a built dir

  - tests/assemblyscript/testsuite/wasm32-wasip1: All wasm and json files copied here


pchickey:
I'm not sure whether there's actually value in the Assemblyscript test suite for p1, but Assemblyscript has made it clear that they would not be supporting p2+, so I'd advocate for putting as little effort into this as possible, since its effectively "done" and won't be ported to any future targets.

loganek (approved):

PR #116, opened 2025-08-21
upload_test_binaries clears testsuite before downloading artifact

wingo:
The previous state was that the build_assemblyscript, build_c, and build_rust workflows upload tests/*/testsuite as an artifact, and then upload_test_binaries runs once for each of {rust,assemblyscript,c}, removing any wasm file in tests/*lang*/testsuite, downloading the artifact, unpacking it to tests/*lang*/testsuite, and adding and committing to prod/testsuite/base if any file differed.

This PR implements a more simple and robust approach, which is to just have upload_test_binaries remove tests/*lang*/testsuite before unpacking the artifact.


loganek (approved):
LGTM

PR #118, opened 2025-08-22
Rework fd_allocate, file_allocate tests to allow for ENOTSUP at run time

wingo:
The TESTCONFIG machinery is not used by the test runner: there is no facility to pass --env arguments to the adapters.


wingo:
Yaargh, PR split failure; rejigging...

pchickey (approved):

PR #120, opened 2025-08-25
First wasip3 test

wingo:
Depends on #114.  Adds wasm32-wasip3 Rust test directory, with a first Rust-based test.  Test runner not yet updated; requires passing `Wcomponent-model-async=y -Sp3=y` to wasmtime.  Building the wasip3 tests requires a nightly Rust toolchain, so let's use that too for wasip1.

### DIFF
--- a/.github/workflows/compile-tests.yml
+++ b/.github/workflows/compile-tests.yml
@@ -47,21 +47,17 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v3
 
-      - name: Install wasm32-wasip1 target for Rust
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          toolchain: stable
-          target: wasm32-wasip1
-
-      - name: Install wasm32-wasip2 target for Rust
+      - name: Install nightly wasm32-wasip2 target for Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
           toolchain: nightly
           target: wasm32-wasip2
 
-      - name: Set default Rust toolchain to stable
-        run: |
-          rustup default stable
+      - name: Install stable wasm32-wasip1 target for Rust
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-wasip1
 
       - name: Build tests
         working-directory: tests/rust


tests/rust/wasm32-wasip3/src/bin/monotonic-clock.rs
```diff
@@ -0,0 +1,75 @@
...
+    assert!(monotonic_clock::get_resolution() > 0);
+}
+
+struct Component;
+export!(Component);
+impl exports::wasi::cli::run::Guest for Component {
+    async fn run() -> Result<(), ()> {
+        test_wait_for().await;
+        test_wait_until().await;
+        test_resolution();
```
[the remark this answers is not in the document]
wingo: Hmm, I am not so sure:
 * There is no such thing as a negative duration in the spec.
 * wait_for with 0 duration is not specified to immediately return.
 * wait_for(now()) may block indefinitely if now() wraps around.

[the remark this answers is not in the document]
rvolosatovs: Looks like we need to update the spec

These have been the expectations we had in p2 as well https://github.com/bytecodealliance/wasmtime/blob/e767c56b824e5ce8947997b052859e050419d35b/crates/test-programs/src/bin/preview2_sleep.rs

> * There is no such thing as a negative duration in the spec.

Rather than `wait_for` negative duration I should have said `wait_until` with an instant in the past

> * wait_until(now()) may block indefinitely if now() wraps around.

I'm not sure I agree.
- Assume time `T_start`
- `now()` is called, returns `T_1`. `T_1 < T_start`, because the clock has wrapped around since `T_start`
- `wait_for` is called with `T_1`
- `wait_for` returns immediately, since time is now `T_2` and `T_2 > T_1`

I guess this is again a question of waiting for an instant in the past.

I strongly disagree, in fact, that waiting for an instant in the past should block indefinitely, since due to non-determinism it seems awfully easy to end up in a deadlock this way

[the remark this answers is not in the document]
lann: There are some slightly stale spec changes proposed in https://github.com/WebAssembly/wasi-clocks/pull/71

wingo:
* Rebased on #114 to resolve conflicts.
 * No longer assumes that `monotonic_clock::now()` wraps; tests the opposite.
 * Add some wait_for(0), wait_until(0), etc tests.

wingo:
Rebased to remove dependent branches from the diff

.github/workflows/compile-tests.yml
```diff
@@ -50,8 +50,15 @@ jobs:
       - name: Install wasm32-wasip1 target for Rust
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
+          toolchain: nightly
```
loganek, line 53: that shouldn't be needed for wasip1, it's already tier2 in Rust which makes it available in the stable toolchain.

tests/rust/build.py
```diff
@@ -21,7 +21,7 @@
     args.verbose = True
 
 SYSTEMS = ['wasm32']
-VERSIONS = ['wasip1'] # + ['wasip2', 'wasip3']
```
loganek, line 24: did you mean wasip2? I see in the compile-tests.yml workflow you configure wasip2 target, not wasip3

tests/rust/wasm32-wasip3/wit/deps/wasi-cli-0.3.0-rc-2025-08-15/package.wit
```diff
@@ -0,0 +1,202 @@
+package wasi:cli@0.3.0-rc-2025-08-15;
```
loganek, line 1: I'm not a wit expert, but can we, instead of copying the files, pull them as a submodule from the original repo: https://github.com/WebAssembly/wasi-cli/ ?

tests/rust/build.py
loganek: I think this is quite confusing and perhaps requires a bit of re-thinking how we define test cases overall.
Perhaps in the test suite metadata we should specify the minimum / maximum version of the WASI version that the test is compatible with, and then the test runner can select an appropriate set of tests to run based on parameters?

tests/rust/wasm32-wasip3/src/bin/monotonic-clock.rs
```diff
@@ -0,0 +1,81 @@
...
+    dur
+}
+
+async fn test_wait_for() {
+    let start = monotonic_clock::now();
+    monotonic_clock::wait_for(1 * MILLISECOND).await;
+    let end = monotonic_clock::now();
+    assert!(compute_duration(start, end) >= 1 * MILLISECOND);
+
+    monotonic_clock::wait_for(0).await;
```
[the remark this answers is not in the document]
wingo: Hmm, where is that specified?

tests/rust/build.py
wingo: Agreed with regards to having a shortcut to run e.g. "just the wasip1 tests".  But from a directory structure / test suite definition perspective, I don't know how to be more clear: the test sources are in `tests/rust/wasm32-wasip3`, and the binaries get written to `tests/rust/testsuite/wasm32-wasip3`.  They test wasip3 APIs.  That they are compiled with a toolchain named "wasip2" is a temporary detail that will go away once the wasip3 release is out.

tests/rust/wasm32-wasip3/src/bin/monotonic-clock.rs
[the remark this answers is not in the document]
rvolosatovs: https://github.com/WebAssembly/wasi-clocks/pull/97

.github/workflows/compile-tests.yml
wingo: And indeed, having made that change, the wasip1 build is now looking for a nightly toolchain.  Annoying!

tests/rust/build.py
pchickey: @wingo I agree that its clear when you lay it out that way - can you add text to the readme that describes how the tests for wasi<version> are in tests/<lang>/wasi<version>, and a comment here in this build script that says wasip2 is an implementation detail until wasip3 ships in rustc?

rvolosatovs (approved):

tests/rust/build.py
pchickey: No, there's no assumption that tests from wasip(N-1) work in wasip(N) - wasi revisions are where we are allowed to make breaking changes, whether because the prior interfaces were flawed or because new component model features allow us to change. So, when its time to add wasip4 (or wasi 1.0 or etc) we'd probably start by making a copy of the prior tests and then making whatever necessary changes according to what changed in the spec.

pchickey: > I'm not fully convinced to the idea of having copies of tests for each version as I'm worried about the overhead this will cause whenever we have to update the tests.

Our goal should be that, for older revisions of wasi, the tests are not updated except to clarify what the behavior of a certain revision should be. Whether a given clarification applies to more than one revision applies on a case by case basis.

@wingo has been working on two engineering activities in parallel, and I think we should distinguish between them. One effort, here, is to create new tests for wasip3. A second effort, is to clarify the spec of wasip1, in e.g. #118 #119 #124. Of those, only #119 has any applicability at all to wasip3, and when it does, it will be rendered with completely different source code because the wasip3 interfaces for expressing fdflags sync are distinct from wasip1. The other two PRs are for aspects of the wasip1 interfaces that we deliberately removed in wasip2, so there is no possible way to share code between those tests.

PR #121, opened 2025-08-27
Fix compile-tests workflow

wingo:
Moving source files out of the built dir effectively removed the built dir, causing merge conflicts in the workflow that rebuilt the wasm tests which were supposed to be in that dir.

Fix by doing a `--strategy ours` merge, then forcibly resetting the tree state to the main branch, then checking out the built files.  This ensures that the tree contains precisely the upstream source code, plus our built files, while keeping the parent tree.


wingo:
Hmm, need to --no-commit the merge I think...  apologies for the noise :/

PR #124, opened 2025-08-28
Disable some tests if rights aren't supported

alexcrichton:
Disable the `fd_fdstat_set_rights` tests entirely if rights aren't supported by the host and skip some checks in `path_link` if rights aren't supported. This gets these tests passing locally for Wasmtime, for example.

### DIFF
--- a/tests/rust/wasm32-wasip1/src/bin/fd_fdstat_set_rights.rs
+++ b/tests/rust/wasm32-wasip1/src/bin/fd_fdstat_set_rights.rs
@@ -1,11 +1,17 @@
 use std::{env, process};
 use wasi_tests::{
     assert_errno, create_tmp_dir, drop_rights, fd_get_rights, open_scratch_directory,
+    supports_rights,
 };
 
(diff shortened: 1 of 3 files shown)


wingo:
LGTM!  I had a similar version of this but I think yours is better.

pchickey (approved):

loganek (approved):

loganek:
I'm ok with that change but I wonder if instead of silently skipping the test failure, the runtime maintainers should rather disable those tests. So in the report, instead of showing "PASS", we can clearly indicate the test was skipped.

alexcrichton:
To the extent that this repository serves as a dashboard for users to review I personally think the behavior should be codified in the test itself. The WASIp1 APIs themselves are so under-specified there's no canonical behavior to rely on so this test suite seems to be retro-actively trying to codify what the real-world differences are between runtimes and some behaviors of the APIs. Given that as well I don't think it'd be useful to have an exclusion in a remote location as opposed to here.

Issue #111, opened 2025-08-15
Update pwrite expected behavior for files opened in append mode


wingo:
In https://github.com/WebAssembly/wasi-testsuite/pull/95, a test was added that `pwrite` implements POSIX semantics (https://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html).  Note that Linux deviates from POSIX, ignoring the "offset" argument for append-only files, instead always writing at the end: https://linux.die.net/man/2/pwrite.  Wasmtime used to paper over this Linux behavior, choosing POSIX instead, but in https://github.com/bytecodealliance/wasmtime/pull/8823 decided to choose the Linux behavior over POSIX, to align with Linux native, as well as WAMR and WasmEdge: https://github.com/bytecodealliance/wasmtime/issues/8817.  The justification was that this part of WASI was under-specified, and I think that Linux developers expected Linux pwrite behavior.  In the meantime, WasmEdge actually changed their behavior to match wasmtime's old behavior; https://github.com/WasmEdge/WasmEdge/issues/3062.

The current status is that the [pwrite-with-append](https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/c/testsuite/pwrite-with-append.c) test currently fails on wasmtime, wazero, wamr, while passing on wasmedge and pywasm-runtime.

IMO either option is fine, but it would be good if wasmedge and wasmtime people could come to agreement on this issue :)

mohanson:
Glad to see a consensus on this issue.

I encountered this issue before, and due to the introduction of this test case https://github.com/WebAssembly/wasi-testsuite/pull/95, I chose to change the pywasm implementation by: https://github.com/mohanson/pywasm/commit/0275285ce46cd4478f249e0492481841a396adb3

I will update pywasm according to the majority rules.

vados-cosmonic:
It's been more than a few days :)

PR #125, opened 2025-08-28
Accept either POSIX or Linux behavior in `pwrite-with-append.c`

alexcrichton:
WASIp1 does not specify which is correct, so accept either which runtimes. There is no current consensus amongst runtimes which is correct.


wingo:
Related to https://github.com/WebAssembly/wasi-testsuite/issues/111; I thought there was a consensus emerging for one or the other?  Dunno tho.

PR #131, opened 2025-09-15
wasi:filesystem@0.3.0-rc-2025-09-16: Add tests for flags/type APIs

wingo:


pchickey (approved):
Only one suggestion from me, maybe you already thought of this and there is a good reason it doesn't work.

PR #138, opened 2025-09-16
wasi:filesystem@0.3.0-rc-2025-09-16: Add tests for is-same-object

wingo:


pchickey (approved):
I expect the most significant features this intersects with are hard links and symbolic links, can you add uses of those?

wingo:
> I expect the most significant features this intersects with are hard links and symbolic links, can you add uses of those?

The funny thing is that we can't test symlinks: we can't `open` a symlink itself, only follow it or `readlink` it.  But I will add some hard-link tests.

Issue #127, opened 2025-08-28
No documented way in README to edit a test and run it


alexcrichton:
Reading over various bits of documentation I had to dive pretty deep to figure out that there wasn't any command/setup to both  build a test and run it. While there's documentation and scripts to do both nothing combines the two. Personally I think this should be documented in the README because any changes to the tests probably wants to be accompanied with both building the test and running the test (and this [isn't currently done in CI either](https://github.com/WebAssembly/wasi-testsuite/issues/126))

PR #149, opened 2025-09-18
Improve test runner ergonomics

wingo:
Add `./run-tests` executable that doesn't require arguments, automatically detecting available runtimes and tests.

Simplify `README.md` accordingly.

Console test case reporter more concise and useful:
 - Single line to indicate total results
 - One copy-pastable command line per failure
 - Requires refactors to record command lines, remove cwd change

Allow testing multiple engines in one run.
 - Required changing `TEST_RUNTIME_EXE` environment variable convention to one env var per executable
 - Requires refactor to collect which runtime ran a testcase

### DIFF
--- a/run-tests
+++ b/run-tests
@@ -35,6 +35,12 @@ parser.add_argument(
     default=False,
     help="Disables color for console output reporter.",
 )
+parser.add_argument(
+    "--verbose",
+    action="store_true",
+    default=False,
+    help="Print more information about test results.",
+)
 
 def find_test_dirs(root):
     test_dirs = []
--- a/test-runner/wasi_test_runner/__main__.py
+++ b/test-runner/wasi_test_runner/__main__.py
@@ -49,7 +49,8 @@ def main() -> int:
 
     options = parser.parse_args()
 
-    reporters: List[TestReporter] = [ConsoleTestReporter(not options.disable_colors)]
+    reporters: List[TestReporter] = [ConsoleTestReporter(not options.disable_colors,
+                                                         verbose=True)]
     if options.json_output_location:
         reporters.append(JSONTestReporter(options.json_output_location))
 
--- a/test-runner/wasi_test_runner/reporters/console.py
+++ b/test-runner/wasi_test_runner/reporters/console.py
@@ -14,18 +14,26 @@ class ConsoleTestReporter(TestReporter):
     _SKIP_COLOR: str = Fore.LIGHTBLACK_EX
     _RESET_COLOR: str = Fore.RESET
 
-    def __init__(self, colored: bool = True) -> None:
+    def __init__(self, colored: bool = True, verbose: bool = False) -> None:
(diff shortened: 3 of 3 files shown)


wingo:
So.  Before, you would run this:
```
TEST_RUNTIME_EXE="$HOME/src/wasip3/wasmtime/target/release/wasmtime -Wcomponent-model-async=y -Sp3=y" \
  python3 test-runner/wasi_test_runner.py \
    -t tests/assemblyscript/testsuite/wasm32-wasip1 \
       tests/c/testsuite/wasm32-wasip1 \
       tests/rust/testsuite/wasm32-wasip{1,3} \
    -r adapters/wasmtime.py
```
And you would get output like this: [before.txt](https://github.com/user-attachments/files/22406339/before.txt)

After this change, you would have to specify the Wasmtime runtime via the `WASMTIME` variable:
```
WASMTIME="$HOME/src/wasip3/wasmtime/target/release/wasmtime -Wcomponent-model-async=y -Sp3=y" \
  python3 test-runner/wasi_test_runner.py \
    -t tests/assemblyscript/testsuite/wasm32-wasip1 \
       tests/c/testsuite/wasm32-wasip1 \
       tests/rust/testsuite/wasm32-wasip{1,3} \
    -r adapters/wasmtime.py
```
And you get similar output, though there are section headings printed as we run tests, and the synopsis is shorter and shows the commands to re-run: [after-wasi-test-runner.txt](https://github.com/user-attachments/files/22406377/after-wasi-test-runner.txt)

But the neat thing is that after, you can just run this:
```
WASMTIME="$HOME/src/wasip3/wasmtime/target/release/wasmtime -Wcomponent-model-async=y -Sp3=y" \
  ./run-tests
```
That gives you similar output, though it prints some info indicating how it detects runtimes in the beginning: 
[after-run-tests.txt](https://github.com/user-attachments/files/22406396/after-run-tests.txt)

In future I would like for `run-tests` to be more terse as it goes, perhaps printing just a dot per test or something, so that everything stays on a page.  We can add `--verbose` or something to print details as we go, for CI porpoises.

WDYT @alexcrichton @pchickey @loganek ?

In future I would like to:
 - Remove the fork/exec around invoking adapters: the adapter should return an argv directly, that way re-invoking a failing test can be done directly instead of popping through python
 - Make `run-tests` more terse
 - Possibly include `build-tests` or something (#127)
 - Add a wasi version to `manifest.json`, so that the runtime adapters can choose different argv for different wasi versions
 - Add a jco adapter
 - Add CI to publish test status; can be a dynamic page that pulls the nightly json data directly from the repo
 - Run tests against some engine as part of CI

alexcrichton:
That looks great to me! I like having copy/pastable commands at the end and I also like the aspiration of "here's the command for running the runtime directly" to get printed eventually as that's exactly what I'd be looking for when debugging things.

PR #154, opened 2025-09-22
Fix build.py for Rust tests

wingo:
Incredibly, Python's `shutil.copytree` doesn't do what it says when copying a tree containing a symlink over the same tree containing the same symlink.  Therefore, we remove the directory first before copying over it, and copy over it at most once per build.


wingo:
Merging so that CI starts working again...

PR #155, opened 2025-09-22
Move support_rights to be a config option

andreaTP:
Since you re-started updating the "release branches", in Chicory we picked up this change: https://github.com/WebAssembly/wasi-testsuite/commit/92c0545ae4044ea6ed370589e4007ac769e182ec

That breaks the integration tests in Chicory since `fd_fdstat_set_rights` is fully not implemented and throws an exception.
If I understand this repository correctly(it was not me doing the initial integration, catching up now), a "runtime agnostic" of disabling tests needs to go through the `TESTCONFIG`.

Fixes: https://github.com/dylibso/chicory/issues/1042


andreaTP:
cc. @wingo :pray:

andreaTP:
Great feedback!

I'll give a try to your suggestion as it should work and call it a day :+1:

PR #156, opened 2025-09-24
Add adapter for pywasm

wingo:
Also add run-pywasm executable script, as pywasm doesn't have one. Update workflow appropriately.  Cc @mohanson.


mohanson (commented):
LGTM

PR #161, opened 2025-09-24
Change daily test workflow to use run-tests

wingo:
In this way we will automatically discover test dirs.  Will need to add some skipping though!


wingo:
With the skipping from #162 I think we are good to merge here.

PR #164, opened 2025-09-25
Add doc on how to write new tests

### LINKED ISSUE #127: No documented way in README to edit a test and run it

wingo:
Fixes #127.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,21 +1,21 @@
 # `wasi-testsuite`
 
-This repository contains tests for WebAssembly System Interface (WASI) and a test executor for running WASI tests against a selected WebAssembly runtime.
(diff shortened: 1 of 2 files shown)


pchickey (approved):
Thanks @wingo this is very appreciated!

PR #168, opened 2025-10-02
Add test for creating HTTP requests

wingo:
Doesn't actually perform the request yet.


wingo:
Just a rebase and update to WASI 2026-02-09.  Had to make one change though: it seems that the http crate allows raw UTF-8 now for codepoints greater than 128: https://github.com/WebAssembly/WASI/issues/791#issuecomment-3934985478.

Issue #181, opened 2025-10-22
Upstream Wasmtime's WASI 0.3 tests


yoshuawuyts:
This is a sibling issue of https://github.com/WebAssembly/component-model/issues/571. Wasmtime currently has its own set of WASI 0.3 tests written in Rust. With an eye on portability, we should upstream these tests from Wasmtime into the wasi-testsuite repo. That way runtimes like Jco will no longer have to rely on Wasmtime for test coverage, but can directly rely on `wasi-testsuite` instead.

I expect the upstreaming to occur in roughly three phases:

1. Upstream a single test to ensure I'm doing things right.
2. Upstream the remaining tests in bulk PRs to migrate everything over from Wasmtime.
3. Refactor the now checked-in tests against the prior tests to reduce the number of duplicate tests.

With this I hope we can accelerate Jco's transition away from relying on Wasmtime's tests suite, and move it to rely on this project instead. Once we finish migrating tests over, we should be to delete the tests from Wasmtime soon after as well. Thanks!

## Tests

_There are a total of 39 test programs. The test program sources can be found [here](https://github.com/bytecodealliance/wasmtime/tree/710ba4bbf33e02f681eeada012ffa081b8d20588/crates/test-programs/src/bin), and their invocations as part of Wasmtime's test suite can be found [here](https://github.com/bytecodealliance/wasmtime/blob/710ba4bbf33e02f681eeada012ffa081b8d20588/tests/all/cli_tests.rs) and [here](https://github.com/bytecodealliance/wasmtime/blob/710ba4bbf33e02f681eeada012ffa081b8d20588/crates/wasi/tests/all/p3/mod.rs)._

- [ ] `p3_api_proxy.rs`
- [ ] `p3_cli_hello_stdout.rs`
- [ ] `p3_cli_much_stdout.rs`
- [ ] `p3_cli_serve_hello_world.rs`
- [ ] `p3_cli.rs`
- [x] `p3_clocks_sleep.rs`
- [ ] `p3_file_write.rs`
- [ ] `p3_filesystem_file_read_write.rs`
- [ ] `p3_http_echo.rs`
- [ ] `p3_http_middleware_with_chain.rs`
- [ ] `p3_http_middleware.rs`
- [ ] `p3_http_outbound_request_content_length.rs`
- [ ] `p3_http_outbound_request_get.rs`
- [ ] `p3_http_outbound_request_invalid_dnsname.rs`
[...]

PR #182, opened 2025-10-22
Upstream clocks tests from wasmtime

yoshuawuyts:
This upstreams the [WASI clocks tests](https://github.com/bytecodealliance/wasmtime/blob/bd7b59dad0923097760eb73103582252fcd2f408/crates/test-programs/src/bin/p3_clocks_sleep.rs) from Wasmtime. This makes progress towards #181, porting a single test to make sure I'm doing things right. Once we land this, I'd like to port the remainder of the tests in bulk.

As I mentioned in #181 as well: I expect there to be some overlap between this test and the existing WASI 0.3 clocks test. Ideally what I'd like to do is finish porting all tests over first, and then move to refactor/de-duplicate some of the tests. My hope is that with that we can unblock Jco from moving onto `wasi-testsuite` sooner, delete the WASI 0.3 tests from Wasmtime, and focus more of the effort on updating the shared testsuite here. Thanks!

### DIFF
--- a/tests/rust/wasm32-wasip3/src/bin/clocks-sleep.rs
+++ b/tests/rust/wasm32-wasip3/src/bin/clocks-sleep.rs
@@ -1,0 +1,65 @@
+extern crate wit_bindgen;
+
+wit_bindgen::generate!({
+    inline: r"
+  package test:test;
+
+  world test {
+      include wasi:clocks/imports@0.3.0-rc-2025-09-16;
+      include wasi:cli/command@0.3.0-rc-2025-09-16;
+  }
+",
+    // Work around https://github.com/bytecodealliance/wasm-tools/issues/2285.
+    features:["clocks-timezone"],
+    generate_all
+});
+
+use core::future::Future;
+use core::pin::pin;
+use core::task::{Context, Poll, Waker};
+use wasi::clocks::monotonic_clock;
+
+struct Component;
+export!(Component);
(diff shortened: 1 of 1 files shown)


pchickey (approved):

wingo:
I took the liberty of rebasing the branch and adding a commit to update to the latest WASIp3.  WDYT @yoshuawuyts ?

tests/rust/wasm32-wasip3/src/bin/clocks-sleep.rs
```diff
@@ -0,0 +1,79 @@
...
+
+  world test {
+      include wasi:clocks/imports@0.3.0-rc-2026-02-09;
+      include wasi:cli/command@0.3.0-rc-2026-02-09;
+  }
+",
+    // Work around https://github.com/bytecodealliance/wasm-tools/issues/2285.
+    features:["clocks-timezone"],
+    generate_all
+});
```
saulecabrera, line 15: (Drive-by) I introduced https://github.com/WebAssembly/wasi-testsuite/blob/main/tests/rust/wasm32-wasip3/src/clocks.rs, to avoid repeating the same pattern on each test, perhaps we could make use of that module here as well?

vados-cosmonic:
These tests are pretty old, and the current repo has updated tests (and is running them w/ the new setup), so I think we are covered now on clocks tests!

PR #183, opened 2025-12-01
Update Rust wasip3 toolchain

wingo:
Now that nightly has a wasip3 target, let's use it.


wingo:
Yaargh!  Ah well, I guess I'll mark this PR as a WIP for the time being.

vados-cosmonic:
The upstream wasi toolchain *has* been integrated at this point and is runnable (and running in CI) in the repo, so I think we should be good here!

PR #220, opened 2026-02-23
Add new HTTP end-to-end test

wingo:
Had to update wit-bindgen to git to fix https://github.com/bytecodealliance/wit-bindgen/pull/1547.

### DIFF
--- a/test-runner/wasi_test_runner/test_case.py
+++ b/test-runner/wasi_test_runner/test_case.py
@@ -1,5 +1,6 @@
 import logging
 import json
+import signal
 from pathlib import Path
 from enum import Enum, StrEnum, auto
 from typing import List, NamedTuple, TypeVar, Type, Dict, Any, Set, Tuple
@@ -10,7 +11,8 @@ CONFIG_KEYS = {"operations", "proposals"}
 
 
 # Supported operations
-SUPPORTED_OPERATIONS = {"run", "wait", "read", "write", "connect", "send", "recv"}
+SUPPORTED_OPERATIONS = {"run", "wait", "read", "write", "connect",
+                        "send", "recv", "request", "kill"}
 
 
 class WasiVersion(StrEnum):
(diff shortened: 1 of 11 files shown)


wingo:
The wasm component is fine and runs under wasmtime via:
```
wasmtime serve -Wcomponent-model-async -Sp3,http,cli tests/rust/testsuite/wasm32-wasip3/http-service.wasm
```

Need to extend test runner to make http requests.  Given that wasmtime adds on other nondeterministic headers (e.g. `date`), and that this is expected and fine, we should probably use an http library from python to test, and extend the vocabulary of the test suite specifications accordingly.

wingo:
We can add more end-to-end tests within the json spec, this is just a first thing so that we can finally check off this "end-to-end" box :)

test-runner/wasi_test_runner/test_case.py
```diff
@@ -166,14 +168,104 @@ def from_config(cls: Type[C], config: Dict[str, Any]) -> C:
...
+Resp = TypeVar("Resp", bound="Response")
+
+
+class Response(NamedTuple):
+    status: int
+    headers: Dict[str, str]
+    body: str
+
+    @classmethod
+    def from_config(cls: Type[Resp], config: Dict[str, Any]) -> Resp:
```
saulecabrera, line 180: Should we add a unit test for this method?

test-runner/wasi_test_runner/test_case.py
```diff
@@ -196,8 +288,9 @@ def from_file(cls: Type[T], config_file: str) -> T:
             if dict_config.get("operations") is not None:
                 operations = cls._operations_from_config(test_config_path, dict_config.get("operations"))
 
-            proposals = []
-            if dict_config.get("proposals") is not None:
+            if dict_config.get("proposals") is None:
+                proposals = _infer_proposals_from_operations(operations)
+            else:
```
saulecabrera, line 293: I _think_ I understand the rationale for doing this, however, is there a risk of introducing a mismatch of specified proposals vs non-specified vs derived? FWIW, I remember that in the original design I was heavily inclined to explicitly stating the target proposals. Could we perhaps validate that if proposals are specified, match the derived ones?

tests/rust/wasm32-wasip3/src/bin/http-service.rs
```diff
@@ -0,0 +1,69 @@
...
+
+  world test {
+      include wasi:http/service@0.3.0-rc-2026-02-09;
+  }
+",
+    additional_derives: [PartialEq, Eq, Hash, Clone],
+    features:["clocks-timezone"],
+    generate_all
+});
+
```
saulecabrera: Just a note that once this is rebased on top of main, we can use the `http` module.

saulecabrera (commented):
LGTM! Some questions inline.

test-runner/wasi_test_runner/test_case.py
wingo: It has one via `test_request_from_config`, as `Request.from_config` calls `Response.from_config`.

test-runner/wasi_test_runner/test_case.py
wingo: Good question.  One comment first, I think really we have "features" rather than "proposals": aspects of the WASI runtime that need to be present for a test to work.  Sometimes this maps to proposals.  In this case the thing that I wrote as `http/service` isn't a proposal really, rather a mode of operation (hence the corresponding wasmtime adapter changes to make it run `wasmtime serve` and enable cli, which is otherwise enabled by default.  I can change the name to "features" in a followup if you like.

But, I hesitated about making the `http/service` feature name visible in the JSON, because it seems like a detail.  Then I realized I could avoid having to say anything if I just automatically inferred what features are needed.

I don't think we can assert however that inferred proposals match explicitly specified ones.  The need for `http` or `sockets` can't be inferred from the set of ops, unlike `http/service`.

wingo:
Updated branch uses http module, adds more tests, fixes support for non-GET methods.

test-runner/wasi_test_runner/test_case.py
saulecabrera: You're right, I missed this one!

test-runner/wasi_test_runner/test_case.py
saulecabrera: > I don't think we can assert however that inferred proposals match explicitly specified ones. The need for http or sockets can't be inferred from the set of ops, unlike http/service.

The `http/service` detail is indeed a good point and I agree, it's a detail and we can infer it, especially if we don't want to expose it as a configuration concept, however I see in the [current test configuration](https://github.com/WebAssembly/wasi-testsuite/blob/860e48c6ba23f19062d13273f1c90ab2bbae1224/tests/rust/wasm32-wasip3/src/bin/http-service.json#L2):

> "proposals": ["http", "http/service"],


Should we perhaps remove it then?

Particularly for inference: in `_infer_proposals_from_operations` we are also inferring `sockets`, which is what I'm not sure a path that we should follow. Maybe we could consider removing the inference of sockets, and focus solely on inferring `http/service`?

saulecabrera: We discussed this offline, we'll tackle it in a follow-up.

saulecabrera (approved):

PR #221, opened 2026-02-24
Refactor test runner

wingo:
When adding HTTP support to the runner, I got a bit distracted about code duplication, and how we had so much state in the run_test method, then realized that all the op interpretation isn't the responsibility of the runtime adapter, rather of the test case... and ended up with this big honkin' patch.


test-runner/wasi_test_runner/test_suite_runner.py
```diff
@@ -25,12 +28,213 @@ class Manifest(NamedTuple):
...
+    def do_write(self, write: Write) -> None: pass
+    def do_wait(self, wait: Wait) -> None: pass
+    def do_connect(self, conn: Connect) -> None: pass
+    def do_send(self, send: Send) -> None: pass
+    def do_recv(self, recv: Recv) -> None: pass
+    def do_cleanup(self, successful: bool) -> None: pass
+
+    def extract(self) -> List[str]:
+        self.run()
+        return self._argv
```
saulecabrera: As far as I can tell, `ArgvExtractor` is only called when skipping the tests; I'm failing to see the benefit of the sub-class approach here vs not having it at all. Is it because of the convenience of the dependency injection that is happening with the `Run` parameter in `do_run`, which holds `run.{args,env,dirs}` needed for `compute_argv`?

Could we consider avoiding the sub-class and fetch `Run` out-of-band instead? e.g., `TestCaseValidator.validate()` could return the test entrypoint, which we would then pass to `compute_argv`, avoiding this class entirely. Thoughts?

saulecabrera (commented):
LGTM, thanks for the refactor. Only one question below.

test-runner/wasi_test_runner/test_suite_runner.py
wingo: Hmm, yes I wasn't sure about this; it was indeed for convenience, but there's nothing very convenient about getting all the inputs ready for `runtime.compute_argv` 🤪 

I have another idea.  Skipped tests don't really need an argv; the UI doesn't surface a way to re-run them.  Let me see if I can remove this entirely.

saulecabrera (approved):

Issue #226, opened 2026-04-02
buck2-based build system


andreiltd:
﻿Hi everyone 👋

I've been working on an experimental fork of this repo that replaces the python build scripts with [buck2](https://buck2.build/), and I'd love to get your thoughts on whether this direction could be useful for the project.

﻿I hope this won't start a "which build system is better" debate, they're all equally bad in their own ways 😄 I ended up using buck2 simply because I already had a collection of wasm component model rules written for other stuff, and starlark's syntax felt like a good fit given that the existing project infrastructure is python.

﻿My primary goal with introducing a build system was to improve the dx for contributors working on wasi-testsuite itself: making it easy to build, test, and iterate without manual toolchain setup. The existing Python test runner and adapter architecture already serves great as a generic framework for downstream embedders.

You can start here to get a feel for how it works:

https://github.com/andreiltd/wasi-testsuite?tab=readme-ov-file#using-buck2

Some of the highlights:

- ﻿Buck2 downloads all toolchains automatically using `http_archive` macro as a build step, for example runtimes, are pinned using bzl [file](https://github.com/andreiltd/wasi-testsuite/blob/main/toolchains/releases.bzl)
- C and Rust tests are compiled using Buck2's built-in polyglot cxx_binary and rust_binary rules, for [example](https://github.com/andreiltd/wasi-testsuite/blob/main/tests/c/BUCK)
- `wasi-sdk`, `wasm-tools`, `nodejs` and `asc` compiler are also downloaded hermetically as build step, see [toolchian definition](https://github.com/andreiltd/wasi-testsuite/blob/main/toolchains/BUCK).
- ﻿Build and test on Linux, macOS, and Windows from a single workflow. Results match upstream.
- ﻿The existing python adapters and runners are reused as-is,
- I went an extra mile and wrote a simple build step `﻿./buck2 build //tests:dist` that produces a self-contained `.tar.gz` with manifest, prebuilt wasm binaries, configs, and fixtures. This can be used instead of ﻿`prod/testsuite-base` branch, after extracting, the runner can just do: `﻿./run-tests wasi-testsuite/ -r adapters/wasmtime.py`. I've used the package to create a rolling GH release of test suite.

﻿I'm genuinely curious whether this direction would be interesting for the project. It's not finished yet, but if there is an interest I can polish it and materialize some PRs. Thanks!   🙏

cc: @saulecabrera @alexcrichton @yoshuawuyts

alexcrichton:
[...]

PR #230, opened 2026-05-29
feat: add initial Buck2 infra for WASI tests

andreiltd:
Introduce a deliberately small Buck2 slice for running few C, Rust, and AssemblyScript wasi Preview 1 tests with Wasmtime through the existing python test runner. This keeps the PR focused on testing the direction and starting discussion, without attempting a broad test-suite refactor just yet. 

See [quickstart](https://github.com/andreiltd/wasi-testsuite/tree/buck2-port-foundation#experimental-buck2-workflow) and https://github.com/WebAssembly/wasi-testsuite/issues/226 for more details on motivation.

### DIFF
--- a/.github/workflows/buck2.yml
+++ b/.github/workflows/buck2.yml
@@ -5,34 +5,46 @@ on:
     branches:
       - main
   pull_request:
-    branches:
-      - '*'
+
+permissions: {}
+
+defaults:
+  run:
+    shell: bash
 
 jobs:
   buck2:
     name: Buck2 smoke test (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-    defaults:
-      run:
-        shell: bash
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
     steps:
       - name: Check out repository code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
 
-      - name: Install Dotslash
+      - name: Install Buck2 helper tools
         uses: taiki-e/install-action@15413b256f995d819248ea62704771a959284285 # v2.79.13
         with:
-          tool: dotslash
+          tool: dotslash,just
 
       - name: Install Rust WASI target
         run: rustup target add wasm32-wasip1
 
+      - name: Configure Buck2 launcher on Windows
+        if: runner.os == 'Windows'
+        run: echo 'BUCK2=dotslash ./buck2' >> "$GITHUB_ENV"
+
       - name: Run Starlark lint
-        run: git ls-files ':(glob)**/BUCK' ':(glob)**/*.bzl' ':(glob)**/*.bxl' | xargs dotslash ./buck2 starlark lint
+        run: just lint-starlark
 
       - name: Run Buck2 tests
-        run: dotslash ./buck2 test //tests/...
+        run: just test-all
--- a/.github/workflows/buck2.yml
+++ b/.github/workflows/buck2.yml
@@ -1,0 +1,38 @@
+name: Buck2
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  buck2:
+    name: Buck2 smoke test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    defaults:
+      run:
+        shell: bash
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install Dotslash
+        uses: taiki-e/install-action@15413b256f995d819248ea62704771a959284285 # v2.79.13
+        with:
+          tool: dotslash
+
+      - name: Install Rust WASI target
+        run: rustup target add wasm32-wasip1
+
+      - name: Run Starlark lint
+        run: git ls-files ':(glob)**/BUCK' ':(glob)**/*.bzl' ':(glob)**/*.bxl' | xargs dotslash ./buck2 starlark lint
+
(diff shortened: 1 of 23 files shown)


.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
+name: Buck2
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
```
vados-cosmonic:
```suggestion
  pull_request:
```
NIT: I think this can be shortened

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+
```
vados-cosmonic, line 14:
```suggestion
permissions: {}
```

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  buck2:
+    name: Buck2 smoke test (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    strategy:
```
vados-cosmonic, line 21:
```suggestion
    permissions:
        contents: read
    strategy:
```

Assuming you take the deny-by-default stuff above!

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
vados-cosmonic: NIT: really like YAML lists here so it's easy to do diffs and add/remove

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+    defaults:
```
vados-cosmonic: NIT: IMO this default could go for the whole file (doesn't matter since there's only one job for now of course!

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+      run:
+        shell: bash
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```
vados-cosmonic, line 30:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
            persist-credentials: false
```

nice security benefi

.github/workflows/buck2.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Install Dotslash
+        uses: taiki-e/install-action@15413b256f995d819248ea62704771a959284285 # v2.79.13
+        with:
+          tool: dotslash
+
+      - name: Install Rust WASI target
+        run: rustup target add wasm32-wasip1
+
+      - name: Run Starlark lint
+        run: git ls-files ':(glob)**/BUCK' ':(glob)**/*.bzl' ':(glob)**/*.bxl' | xargs dotslash ./buck2 starlark lint
```
vados-cosmonic:
```suggestion
        run: just lint-starlark
```

Any difference here? Feels like you could use the [cache acion](https://github.com/taiki-e/cache-cargo-install-action) for both dotslash and just here (or even cargo binstall, not sure if that's installed on runners by default though)

scripts/check.bxl
```diff
@@ -0,0 +1,73 @@
+# Adapted from https://github.com/dtolnay/buck2-rustc-bootstrap/blob/master/scripts/check.bxl
+#
+# Report C/C++ warnings from every dependency of the Wasmtime suite:
```
vados-cosmonic, line 3: These commands feel like great "quickstart" content -- could they be made part of the `justfile` and/or README?

justfile
```diff
@@ -0,0 +1,45 @@
+buck := env_var_or_default("BUCK2", "./buck2")
+
+set shell := ["bash", "-o", "pipefail", "-c"]
+
+default:
```
vados-cosmonic:
```suggestion
@_default:
```

Also of course `@just` is fine too, but I don't know how else you would hide `default` from the output

justfile
```diff
@@ -0,0 +1,45 @@
+buck := env_var_or_default("BUCK2", "./buck2")
+
+set shell := ["bash", "-o", "pipefail", "-c"]
```
vados-cosmonic, line 3: Does this work properly on MacOS? I find myself often having to do something like 

```
shell := if os() == "macos" { "/bin/zsh" } else { "/usr/bin/zsh" }
```

Just to get proper access to `zsh` if I'm assuming it's present

README.md
```diff
@@ -71,6 +71,30 @@ Optionally you can specify test cases to skip with the `--exclude-filter` option
...
+or if you prefer using `just`
+
+``` bash
+just build
+just test
+```
+
+Buck2 fetches the WASI SDK, Node/AssemblyScript, and Wasmtime through the Buck
+toolchain graph, then uses a generic `wasi_test` rule to invoke
+`wasi_test_runner` for each Buck test target.
```
vados-cosmonic, line 104: Mentioned this before, but a few examples like you put in the `.bzl` file would go great here IMO

vados-cosmonic (approved):
Did a quick run through, looks good to me though of course I'm not a Buck2 expert, it seems like most people should be able to get in and get things working relatively simply. 

It might be worth tagging a few things for follow ups (out of scope here IMO):

- updating lib versions (`wasmtime`, `wasi`, wasi-sdk, etc)
- adding a `docs/buck.md` which can go into slightly more detail about structure (though IMO people can mostly search for `BUCK` and figure out what's going on I think.

README.md
andreiltd: I added few commands:

```bash
just lint-starlark
just lint-cxx
just lint-rust
```

justfile
andreiltd: Hm... yeah, I have no way to check this but it seems to work fine in macos CI runner.

vados-cosmonic: Yeah I think that's because the macos runners generally [have `bash` installed on them](https://github.com/actions/runner-images/blob/fdef93069302a54121ddf40ca57b7bec0dac9dc6/images/macos/macos-26-arm64-Readme.md). 


MacOS has `zsh` by default though these days -- maybe most developers will end up downloading `bash` so maybe this will work for most people. That said, let's see if anyone gets bit by this first I guess

ricochet (approved):

PR #232, opened 2026-06-03
feat: move wasi p3 parent symlink setup into tests

andreiltd:
Remove the checked in `fs-tests.dir/parent -> .. symlink` and create the escape symlink at runtime as parent.cleanup in the tests that need it. This keeps the symlink escape coverage while avoiding recursive fixture trees in generated test archives.


vados-cosmonic (approved):
LGTM 🚀 

This also feels like a good time to make that name (`parent`/`parent.cleanup`) a global or variable of some sort.

PR #233, opened 2026-06-03
feat(buck2): buckify all the tests

andreiltd:
This adds wasmtime coverage across all wasi p1 and p3, and package the resulting test data into the dist archive.

The buck2 dependencies are now in:
- third-party/rust
- third-pary/python

BUCK file for Rust dependencies has been generated using: https://github.com/facebookincubator/reindeer

The dist package is now ~49M. The packaging script is using python implementation of archiver, so it is a bit slow. We may think about writing a small rust utility instead - that should be much faster.

The patch is also adding explicit skip files per OS for wasmtime that is consumed by test runner.

The follow up PR will exted matrix to different runtimes. At this point we can remove all the python based build system scripts.


tests/rust/wasm32-wasip3/Cargo.toml
```diff
@@ -4,5 +4,5 @@ version = "0.1.0"
 edition = "2024"
 
 [dependencies]
-wit-bindgen = { git = "https://github.com/bytecodealliance/wit-bindgen.git", rev = "85d10eb33a54191220d34215bbcc4e29cb959884", features = ["async-spawn"] }
-futures = "0.3.31"
+wit-bindgen = { version = "=0.57.1", features = ["async-spawn"] }
```
vados-cosmonic: NIT: Why is this pinned so tightly?

vados-cosmonic (approved):
LGTM 🚀

tests/rust/wasm32-wasip3/Cargo.toml
andreiltd: Good question, I just made it tight as a workaround for reindeer fixup for wit-bindgen:

The build script expects this:

```
DEBUG_OUTPUT_DIR = "$(location :wit-bindgen-rust-macro-0.57-build-script-run[out_dir])"
```

I couldn't figure out how to get the crate + version name here. I can have another look though. That said "0.57" should be fine as well.

vados-cosmonic: Ah thanks for explaining, yeah I think we can keep it simple for as long as possible here... but otherwise maybe [`cargo_metadata`](https://crates.io/crates/cargo_metadata) might be a reasonable way to get at whatever version we are using + a dynamic buck2 target (?)

BUT IMO doing it the simple way for as long as we can is just fine.

PR #235, opened 2026-06-04
feat(buck2): add extra wasi runtime coverage

andreiltd:
Add Buck2 runtime descriptors for Wazero, WasmEdge, and WAMR, including release metadata, adapters, runtime suites, and expectation filters.

This patch skips pywasm because it would mostly be a large expectation file around known runtime limitations: it fails 47/72.

It is now possible to run something like:

```
just test-rust p1 <wamr|wazero|wasmedge|wasmtime>
```

### DIFF
--- a/toolchains/BUCK
+++ b/toolchains/BUCK
@@ -4,8 +4,8 @@ load("@prelude//toolchains:genrule.bzl", "system_genrule_toolchain")
 load("@prelude//toolchains:python.bzl", "remote_python_toolchain")
 load("@prelude//toolchains:remote_test_execution.bzl", "remote_test_execution_toolchain")
 load("@prelude//toolchains:rust.bzl", "system_rust_toolchain")
-load("@root//tools:runtime.bzl", "wasi_runtime")
-load(":runtimes.bzl", "download_wasmtime_runtime")
+load("@root//tools:runtime.bzl", "WAMR_COMPATIBLE_WITH", "wasi_runtime")
+load(":runtimes.bzl", "download_wamr_runtime", "download_wasmedge_runtime", "download_wasmtime_runtime", "download_wazero_runtime")
 load(
     "@wasmono//:defs.bzl",
     "asc_toolchain",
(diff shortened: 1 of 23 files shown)


toolchains/BUCK
```diff
@@ -87,6 +87,13 @@ wasm_tools_toolchain(name = "wasm_tools", distribution = ":wasm_tools_dist", vis
 # Runtime distributions and descriptors. The descriptor is what `wasi_test`
 # consumes, so adding a runtime means adding another descriptor target here.
 download_wasmtime_runtime(name = "wasmtime_dist", version = "45.0.0")
+download_wasmedge_runtime(name = "wasmedge_dist", version = "0.17.0")
+download_wazero_runtime(name = "wazero_dist", version = "1.12.0")
+download_wamr_runtime(
+    name = "wamr_dist",
+    version = "2.4.4",
```
vados-cosmonic: Can we get these versions similar to how we get WAMR_COMPATIBLE_WITH so we can edit in one place and have all the versions change?

toolchains/runtimes.bzl
```diff
@@ -70,8 +83,24 @@ def _download_runtime(name: str, releases, runtime_name: str, version: str):
...
+        **kwargs
     )
 
 def download_wasmtime_runtime(name: str, version: str = "45.0.0"):
     """Download a prebuilt Wasmtime runtime."""
     _download_runtime(name, WASMTIME_RELEASES, "wasmtime", version)
+
+def download_wamr_runtime(
+        name: str,
+        version: str = "2.4.4",
```
vados-cosmonic: Same note here, *if* we can get the default versions specified as a global that is reused here, it would be nice

vados-cosmonic (approved):
LGTM 🚀

toolchains/BUCK
andreiltd: Done!

toolchains/runtimes.bzl
andreiltd: Done!

PR #237, opened 2026-06-08
feat(ci): use buck2 for daily ci tests

vados-cosmonic:
Some refactoring of the `justfile` in here as well, but I've updated the daily runtime validation to run similarly to the new buck2 tests.

### DIFF
--- a/.github/workflows/daily-runtime-validation.yml
+++ b/.github/workflows/daily-runtime-validation.yml
@@ -32,25 +32,16 @@ jobs:
       - name: Install Rust WASI targets
         run: rustup target add wasm32-wasip1 wasm32-wasip2
 
-      - name: Configure Buck2 launcher on Windows
-        if: runner.os == 'Windows'
-        run: |
-          echo 'BUCK2=dotslash ./buck2' >> "$GITHUB_ENV"
-          echo 'MSYS2_ARG_CONV_EXCL=//' >> "$GITHUB_ENV"
-
       - name: Run all tests
+        env:
+          MSYS_NO_PATHCONV: "1"
+          MSYS2_ARG_CONV_EXCL: "*"
         run: just test
 
   other-runtimes:
-    runs-on: ${{ matrix.platform }}
+    runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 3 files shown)


vados-cosmonic:
It looks like GHA is having some issues downloading with 504 right now (both in the cache action, and in the middle of buck2 runs... *with* `persist-credentials: false` removed)...

We maybe should consider committing some amount of the upstream artifacts to the repository, despite the heft, and setting up some sort of checker similar to the daily-test-update that checks for new artifacts and opens a PR for updating.

.github/workflows/daily-runtime-validation.yml
```diff
@@ -4,85 +4,66 @@ on:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          persist-credentials: false
           fetch-depth: 0
-          ref: prod/testsuite-base
 
-      - name: Initialize Python environment
-        uses: actions/setup-python@v4
+      - name: install just
+        uses: taiki-e/cache-cargo-install-action@417450f3c33ee20393705369577571770643d4c7 # v3.0.7
```
andreiltd, line 28: Alternatively we can use: https://github.com/facebook/install-dotslash. It looks like taiki is building dotenv while facebook action installs prebuilt artifact.

vados-cosmonic: Ah sure -- happy to use that instead! The cache cargo install action builds only when it doesn't have a cached version available, but getting dotslash via facebook and always taking the prebuilt is fine with me

vados-cosmonic: Done in https://github.com/WebAssembly/wasi-testsuite/pull/237/commits/45424ed58f2d948906c8cbacddc234108066b069

andreiltd (approved):
LGTM!

vados-cosmonic:
I'm going to try running the tests later on, hopefully when GH is willing to allow some more downloads and the 504s go away :crossed_fingers:

PR #246, opened 2026-06-09
feat(buck2): remove custom python based build system

andreiltd:
This patch removes custom build system and consolidates CI workflows using buck2.


vados-cosmonic (approved):
LGTM :rocket:

alexcrichton:
@andreiltd I think this may have accidentally regressed the [publish-test-binaries workflow](https://github.com/WebAssembly/wasi-testsuite/actions/runs/27219031249/job/80369389916)?

andreiltd:
Hey, yes sorry about that! I was hoping this will fix it: https://github.com/WebAssembly/wasi-testsuite/pull/247 but I have no way to check it since this job is triggered on pushing to main.

The long story short is that [zizmor](https://github.com/zizmorcore/zizmor) doesn't like preserving credentials from checkout job -- so I disabled that in main workflow: https://github.com/WebAssembly/wasi-testsuite/blob/main/.github/workflows/ci.yml#L32

But then plain `git push` doesn't work anymore and I needed this whole dance with authorization header ( see https://github.com/WebAssembly/wasi-testsuite/pull/247). If 247 doesn't work we should just remove the line with preserving credentials and use `git push` again. `¯\_(ツ)_/¯`

alexcrichton:
Oh no worries! Just wanted to flag in case you hadn't seen, but you were way ahead of me...

PR #251, opened 2026-06-10
feat: enable easier single test debugging

vados-cosmonic:


andreiltd (approved):
LGTM!

test-runner/wasi_test_runner/test_suite_runner.py
```diff
@@ -116,12 +116,18 @@ def do_run(self, run: Run) -> None:
             creationflags = 0
             if os.name == "nt":
                 creationflags = getattr(subprocess, "CREATE_NEW_PROCESS_GROUP")
+
+            env = {}
```
andreiltd: Could this be cause of the failures on windows? We are clearing the env that would be normally inherited by `popen`:

> If env is not None, it must be a mapping that defines the environment variables for the new process; these are used instead of the default behavior of inheriting the current process’ environment.

So perhaps the fix is to just use `env = os.environ.copy()` instead of clean env?

vados-cosmonic: yeah that certainly seems plausible -- fixed in https://github.com/WebAssembly/wasi-testsuite/pull/251/commits/e7e444103670ef70134a13e4fe843d2f8e35a0e7, will see if that passes!

PR #253, opened 2026-06-10
feat(buck2): use hermetic rust

andreiltd:
That patch make the build system fully hermetic. The approach for rust hermetic toolchain has been adapted from: https://github.com/rue-language/rue/blob/trunk/toolchains/rust/defs.bzl

It also splits toolchains directory into: rust, cxx, runtimes.


vados-cosmonic (approved):
LGTM 🚀

PR #254, opened 2026-06-10
chore: update Jco to v1.22.0

vados-cosmonic:
Now that [jco v1.22.0](https://www.npmjs.com/package/@bytecodealliance/jco/v/1.22.0) is released, we should update and we should see the stuck windows UDP tests pass.


andreiltd:
Should we remove the udp tests from expectations? It is now skipped.

andreiltd (approved):

vados-cosmonic:
Yeah that's a good idea -- no need to wait for a follow up to address that

PR #260, opened 2026-06-11
wasi 0.3.0

ricochet:
The wasip3 packages are now officially released as 0.3.0!

Replace the release candidate with the v0.3.0 release.

NOTE: wait for wasmtime and jco to vendor 0.3.0 support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /tests/assemblyscript/testsuite/
 __pycache__
 buck-out/
+target/
 
 *~
 \#*\#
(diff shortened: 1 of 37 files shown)


alexcrichton:
Could this actually go ahead and land with adjusted expectations that the tests are skipped/fail on wasmtime/jco? That way the test binaries can get updated and would avoid the chicken-and-egg issue of what to update first perhaps. I ran into this where I'm flagging all wasi tests as "expected to fail" in the wasmtime repo at this time because the version numbers changed.

andreiltd:
I added `expected = fail` to expectations schema, if that's what you prefer to use here instead of skipping tests: https://github.com/WebAssembly/wasi-testsuite/pull/261

alexcrichton (approved):

PR #261, opened 2026-06-12
feat: add expected field to expectations schema

andreiltd:
Allow running tests but accept fail as an outcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ option (the Buck test suites wire these up automatically). Expectation files use
 TOML and support two independent per-test directives:
 
 - `action = "skip"` - don't run the test at all.
-- `expected = "pass"` - run it, and expcet to pass (default can be omitted)
+- `expected = "pass"` - run it, and expect to pass (default can be omitted)
 - `expected = "fail"` - run it, but expect it to fail.
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -65,11 +65,16 @@ overridden by setting corresponding environment variables (`WASMTIME`,
 WASMTIME="wasmtime --wasm-features all" ./run-tests
 ```
 
-Optionally you can specify test cases to skip with the `--exclude-filter` option.
-Expectation files use TOML.
+Optionally you can pass per-runtime expectation files with the `--expectations`
+option (the Buck test suites wire these up automatically). Expectation files use
+TOML and support two independent per-test directives:
+
+- `action = "skip"` - don't run the test at all.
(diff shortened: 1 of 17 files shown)


tools/run_wasi_test.py
```diff
@@ -41,7 +41,7 @@ def main() -> int:
     parser.add_argument("--adapter", required=True)
     parser.add_argument("--config")
     parser.add_argument("--fixture-dir", nargs=2, action="append", default=[])
-    parser.add_argument("--exclude-filter")
+    parser.add_argument("--expectations")
```
ricochet, line 44: I think we need to be careful breaking the CLI here. 

This testsuite's run-tests CLI is a public interface consumed by external runtime CIs (wasmtime, wazero, wamr, jco, and downstream projects). Any of them invoking `--exclude-filter` will break. Consider keeping `--exclude-filter` as a hidden deprecated alias (argparse aliases/a second add_argument mapping to the same dest). The semantics are a strict superset, so an alias is cheap.

test-runner/wasi_test_runner/test_case.py
```diff
@@ -51,6 +51,32 @@ def failed(self) -> bool:
...
+    SKIP = "skip"
+    # Test failed exactly as the expectation file predicted.
+    XFAIL = "xfail"
+    # Test was expected to fail but passed (the expectation is now stale).
+    XPASS = "xpass"
+
+    @property
+    def is_failure(self) -> bool:
+        # Outcomes that turn the suite (and CI) red.
+        return self in (Outcome.FAIL, Outcome.XPASS)
```
ricochet, line 69: Defined and documented but never referenced. The actual exit-code path uses `test_suite.fail_count  > 0` in harness.py:49. Either wire run_all_tests/fail_count to use is_failure or drop the property

README.md
```diff
@@ -65,11 +65,16 @@ overridden by setting corresponding environment variables (`WASMTIME`,
...
 ```
 
-Optionally you can specify test cases to skip with the `--exclude-filter` option.
-Expectation files use TOML.
+Optionally you can pass per-runtime expectation files with the `--expectations`
+option (the Buck test suites wire these up automatically). Expectation files use
+TOML and support two independent per-test directives:
+
+- `action = "skip"` - don't run the test at all.
+- `expected = "pass"` - run it, and expcet to pass (default can be omitted)
```
ricochet:
```suggestion
- `expected = "pass"` - run it, and expect to pass (default can be omitted)
```

vados-cosmonic (approved):

PR #264, opened 2026-06-18
Refactor filesystem tests to have a root directory

alexcrichton:
This commit is a refactor of all filesystem tests and their configuration to have only a single root directory instead of a list of preopened directories. The purpose of this change is to start moving in the direction of "there's just a root filesystem for WASI guests" as opposed to guests having a set of preopened directories they get to choose from. Tests are refactored in how they access preopened directories and search around for them. The JSON configuration of `dirs` is now just `root` and is optionally specified like before. Additionally passing the name of the first preopen as an argument is now no longer necessary so that's removed too.

### DIFF
--- a/adapters/jco.py
+++ b/adapters/jco.py
@@ -3,7 +3,7 @@ import os
 import shlex
 import subprocess
 from pathlib import Path
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Optional
 
 
 # shlex.split() splits according to shell quoting rules.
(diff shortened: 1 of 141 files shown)


andreiltd (approved):

andreiltd:
The `wasmtime` p3 tests are still skipped, should we wait with merging until `wasmtime` is bumped and all tests are passing or you need these changes in `wasmtime`?

alexcrichton:
I tested this locally against a `main`-built wasmtime with the updated WASI versions and everything passed locally. One option is to use the `dev` release of Wasmtime here but that's a moving target so I wouldn't want to inflict that on the CI here

PR #270, opened 2026-07-14
p3(http): Expand tests for http roundtrip

saulecabrera:
This patch introduces `http-service-echo.rs` which is an enhancement to the existing `http-service.rs` test. Particularly, it tests:

* Reading a non-empty inbound body
* Reading inbound headers
* Teaches the python runner and configuration how to send payloads in http requests.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -6,8 +6,10 @@ pkgs.mkShell {
     python3
     python3Packages.pip
     python3Packages.virtualenv
+    buck2
+    dotslash
   ];
-  
+
   shellHook = ''
     # Create and activate virtual environment
     if [ ! -d .venv ]; then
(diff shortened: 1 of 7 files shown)


shell.nix
```diff
@@ -6,8 +6,10 @@ pkgs.mkShell {
     python3
     python3Packages.pip
     python3Packages.virtualenv
+    buck2
+    dotslash
```
saulecabrera, line 10: I added this since it was missing from the switch to buck2. This is meant for local development only and I think I am the only one using nix-shell / nix-develop for dependency management, so when I initially introduced the flake I thought that it might not be worth it to spend CI time on verifying it.

andreiltd: Oh, I missed that, sorry for breaking nix workflow!

andreiltd (approved):
LGTM!

PR #272, opened 2026-07-15
p3(http): Assert request get-{authority,scheme}

saulecabrera:
This commit tests the presence of the authority and scheme in the inbound http request.

The presence/absence of these fields in the inbound request is currently not finalized in the spec[1], hence why this test is expected to fail.

[1]: https://github.com/WebAssembly/WASI/issues/788


andreiltd (approved):

saulecabrera:
I will note, for posterity, that p2, at least in Wasmtime, does derive the authority/scheme from the host, see https://github.com/bytecodealliance/wasmtime/pull/8923

PR #277, opened 2026-07-30
http: Add infrastructure for outbound http requests

saulecabrera:
This commit adds the base infrastructure for outbound http requests; this will enable testing the http client interface.

This patch introduces a new configuration option: `endpoints`, from which the runner sets up an http server with the routing and payload metadata.


saulecabrera:
Please hold off reviewing, I need to add a couple of tests.

saulecabrera:
Ok, should be good now.

andreiltd (approved):

PR #279, opened 2026-08-04
http: Add roundtrip tests for status and method

saulecabrera:
This commit adds proper roundtrip tests for client status and method ensuring that the request reaches the server and response reaches back the client with the appropriate expectations.


saulecabrera:

PR #280, opened 2026-08-04
http: Add client headers tests

saulecabrera:
This builds on top of #279 and adds test to exercise headers from the perspective of the http client.

This commit switches the `/echo` convention into a proper configuration option, called `mode`. `mode`, will configure the action that needs to be taken by the server in terms of response body and headers.


saulecabrera:

PR #281, opened 2026-08-05
http: Cover client.send error cases

saulecabrera:
Builds on top of WebAssembly#280

This commits adds a tests covering two cases for `client.send`:

1. Invalid authority
2. Closed authority

Additionally, in order to enable coverage for the cases above, this
commit augments the configuration by introducing a top level `servers`
configuration key.

This allows creating servers for different purposes, e.g., one that will
be listening for the life span of the test and other that will be closed
immediately, to test closed connection scenarios.


andreiltd (approved):

saulecabrera:

PR #285, opened 2026-08-12
http: Add test for outbound response and request trailers 

saulecabrera:
Builds on top of #284

Adds the necessary machinery to handle outbound request and response
trailers.

This change defaults to including a small trailers parser since the
builtin http library does not include one. The parser is small enough
that it does not justify pulling in an extra dependency.


andreiltd (approved):

saulecabrera:
Merged as part of https://github.com/WebAssembly/wasi-testsuite/pull/286

saulecabrera:
Merged as part of https://github.com/WebAssembly/wasi-testsuite/pull/286

PR #287, opened 2026-08-25
Add adapter for zwasm

chaploud:
[zwasm](https://github.com/zwasm/zwasm) is a from-scratch WebAssembly runtime
written in Zig. An interpreter and a JIT share one WASI preview1
implementation, and both are exercised by the same corpus.

This adds `adapters/zwasm.py` following the module contract in
`doc/adapters.md`. No other file changes.

## The one quirk

zwasm parses options per subcommand rather than globally, so `run` has to be
spliced directly after the executable and ahead of anything the caller put in
`$ZWASM`:

```python
argv += ZWASM[0:1] + ["run"] + ZWASM[1:]
```

That keeps `ZWASM="zwasm --engine jit"` working, which is how I measured both
engines against the same corpus.

## Why only wasm32-wasip1 is declared

zwasm serves preview1 and preview2 in a default build; preview3 is behind a
build option (`-Dwasi=p3`), so a stock binary does not offer it. Declaring
`wasm32-wasip2` would add nothing either — there are no preview2 tests in
`tests/`.

There is also a reason not to declare preview3 even for a build that has it:
guest stdout does not reach a pipe while the guest is still running
([zwasm#255](https://github.com/zwasm/zwasm/issues/255)), which would hang this
suite's interactive tests rather than fail them. Declaring preview1 only keeps
every run here terminating.

## What I checked

Against this repository at `e1f53e0` and zwasm v2.6.0
([`67772ed`](https://github.com/zwasm/zwasm/commit/67772ed72)):

- **With zwasm absent it degrades like every other adapter.** `run-tests` on a
  PATH without it prints `zwasm.py unavailable` and carries on — the missing
  binary surfaces as `UnavailableRuntimeAdapterError`, not as an exception out
  of the adapter. Nothing here can break a run for someone who does not have
  zwasm installed.
- **Discovery and metadata.** `zwasm.py: zwasm 2.6.0`.
- **End to end**, driving zwasm's vendored snapshot of the preview1 corpus
  (72 modules) through `run-tests -r adapters/zwasm.py`: all 72 pass on
  `--engine interp`, on `--engine jit`, and on the default.

  I used that snapshot rather than a Buck-built tree, so the artifacts are not
  this repo's own build. Happy to redo it the other way if that matters for
  review.

Nothing fails, so there is no expectations file to go with this. `pywasm` and
`wizard` are adapter-only in the same way, which is the shape I followed —
adding zwasm to the `wasi_suite` targets in `tests/BUCK` would mean this repo's
CI needs a zwasm binary, which is your call, not something a contribution
should assume. Say the word if it is of interest and I will follow up.

## Small note on the docs

`doc/adapters.md` says the module must define `get_name`, `get_version`, and
`compute_argv`. `RuntimeAdapter.__init__` also calls `get_wasi_versions` and
`get_wasi_worlds`, and an adapter missing either raises `AttributeError`, which
is not in the constructor's catch list — so it fails to load rather than
reporting as unavailable. Happy to send a separate PR for that line if it is
useful.


andreiltd:
Hey, a bit slow here, sorry about that. The patch looks reasonable, the only request I have is to add `zwasm` to the runtime list in  `doc/adapters.md`. 

I also wouldn't mind adding `zwasm` to `test-extra-runtimes` target -- it currently run linux only binaries. It would be a good follow up to prove this repository integration.

andreiltd:
> Happy to send a separate PR for that line if it is useful.

That would be great, yes!

andreiltd (approved):

chaploud:
Added zwasm to the runtime list in ad4f90d. The `doc/adapters.md` contract fix follows in its own PR.

PR #288, opened 2026-09-03
doc: list every function an adapter module must define

chaploud:
`doc/adapters.md` names three functions as the adapter contract, but
`RuntimeAdapter.__init__` also calls `get_wasi_versions` and
`get_wasi_worlds`. An adapter written to the documented list raises
`AttributeError` on load, and that is not in the constructor's catch list
(`subprocess.CalledProcessError`, `FileNotFoundError`), so it fails to load
rather than reporting as unavailable.

Follow-up to #287, as agreed there.


andreiltd (approved):
Thanks!

